PR #8, opened 2019-02-21
change code structure and try to clean wrapper definitions

MaggieQi:


AnnService/src/Core/CommonDataStructure.cpp
```diff
@@ -101,13 +101,27 @@ ByteArray::Data() const
 }
 
 
+void
+ByteArray::SetData(std::uint8_t* p_array)
```
AlphardWang: Suggest not to set members separately, which may mess up the data.

PR #23, opened 2019-05-16
Fix PEP8 voilations in GettingStarted

richardARPANET:
The getting started guide python code has multiple [PEP8](https://www.python.org/dev/peps/pep-0008/) errors in it, which I've fixed.


richardARPANET:
Code review please @MaggieQi  @AlphardWang

PR #33, opened 2019-05-17
Minor rework of SpinLock class for making it more efficient on Unlock()

noSTALKER:
+ atomic_flag is initialized using the recommended ATOMIC_FLAG_INIT
+ Unlock() function clears the atomic_flag using more efficient std::memory_order_acquire
+ Unlock() and Lock() methods are now both noexcept
+ Removed unnecessary include files

### DIFF
--- a/AnnService/inc/Helper/Concurrent.h
+++ b/AnnService/inc/Helper/Concurrent.h
@@ -20,19 +20,19 @@ namespace Concurrent
 class SpinLock
 {
 public:
-	SpinLock() = default;
+    SpinLock() = default;
 
-	void Lock() noexcept 
-	{
-		while (m_lock.test_and_set(std::memory_order_acquire))
-		{
-		}
-	}
+    void Lock() noexcept 
+    {
+        while (m_lock.test_and_set(std::memory_order_acquire))
+        {
+        }
+    }
 
-	void Unlock() noexcept 
-	{
-		m_lock.clear(std::memory_order_release);
-	}
+    void Unlock() noexcept 
+    {
+        m_lock.clear(std::memory_order_release);
+    }
 
     SpinLock(const SpinLock&) = delete;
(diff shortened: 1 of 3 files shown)


noSTALKER:
Just a query for future. I haven't seen `noexcept `keyword usage much in the code base. Is it intentional or i can include it in future pull request?

noSTALKER:
Also the the functions are name `Lock()` and `UnLock() so it can't be used in a code section safely which can throw exception with `std::scoped_lock()` or `std::lock_guard()` as it requires function names to be `lock()` and `unlock()`

AlphardWang:
> Just a query for future. I haven't seen `noexcept `keyword usage much in the code base. Is it intentional or i can include it in future pull request?

Currently we don't have intention to use or not to use noexcept. Please feel free to use it in your PR. We will do refactoring and add it into readme if it becomes an explicit requirement in the future.

AnnService/inc/Helper/Concurrent.h
```diff
@@ -21,17 +20,25 @@ namespace Concurrent
...
+	{
+		m_lock.clear(std::memory_order_release);
+	}
 
     SpinLock(const SpinLock&) = delete;
     SpinLock& operator = (const SpinLock&) = delete;
 
 private:
-    std::atomic_flag m_lock;
+    std::atomic_flag m_lock = ATOMIC_FLAG_INIT;
```
AlphardWang: Personally I don't recommend to use default member initializer since it may become confused and hard to read when there are multiple data members or constructors. But it should be fine here since it is the only data member and all implementation is in same file now.

noSTALKER: This is the recommended way of in initializing `std::atomic_flag`, that is why i didn't used it in initializer list as it requires copy initialization using the `ATOMIC_FLAG_INIT`

PR #51, opened 2019-05-24
Fix assignment operator issues in ByteArray

noSTALKER:
+ Fix assignment operator issues in ByteArray
+ Add noexcept qualifier to functions

### DIFF
--- a/AnnService/src/Core/CommonDataStructure.cpp
+++ b/AnnService/src/Core/CommonDataStructure.cpp
@@ -3,28 +3,29 @@
 
 #include "inc/Core/CommonDataStructure.h"
 
-using namespace SPTAG;
+namespace SPTAG
+{ 
 
 const ByteArray ByteArray::c_empty;
 
-ByteArray::ByteArray()
-    : m_data(nullptr),
-    m_length(0)
+ByteArray::ByteArray() noexcept
+    : m_data(nullptr)
+    , m_length(0)
 {
 }
 
-
-ByteArray::ByteArray(ByteArray&& p_right)
-    : m_data(p_right.m_data),
-    m_length(p_right.m_length),
-    m_dataHolder(std::move(p_right.m_dataHolder))
+ByteArray::ByteArray(ByteArray&& p_right) noexcept
+    : m_data(p_right.m_data)
+    , m_length(p_right.m_length)
+    , m_dataHolder(std::move(p_right.m_dataHolder))
 {
+    p_right.m_data = nullptr;
+    p_right.m_length = 0;
 }
-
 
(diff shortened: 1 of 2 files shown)


AnnService/src/Core/CommonDataStructure.cpp
```diff
@@ -3,130 +3,102 @@
...
+{ 
 
 const ByteArray ByteArray::c_empty;
 
-ByteArray::ByteArray()
-    : m_data(nullptr),
-    m_length(0)
+ByteArray::ByteArray() noexcept
+    : m_data(nullptr)
+    , m_length(0)
```
AlphardWang: please keep original formatting.

AnnService/src/Core/CommonDataStructure.cpp
```diff
@@ -3,130 +3,102 @@
...
 
-
 ByteArray&
-ByteArray::operator= (const ByteArray& p_right)
+ByteArray::operator=(const ByteArray& p_right) noexcept
 {
-    m_data = p_right.m_data;
-    m_length = p_right.m_length;
-    m_dataHolder = p_right.m_dataHolder;
+    if (this != std::addressof(p_right))
```
AlphardWang: I think shared_ptr has already handled such case? Not sure if this is necessary.

AnnService/src/Core/CommonDataStructure.cpp
noSTALKER: I have reverted the changes done in member initializer lists

AnnService/src/Core/CommonDataStructure.cpp
noSTALKER: I haven't checked yet but it is just a recommended standard check in both copy constructor and move constructors

AlphardWang: I take some investigation and think this check is not necessary for std::shared_ptr and most other stl containers (it seems a popular implementation for shared_ptr assign function is to create a temp var and then swap it with `this`). So please remove this check.

noSTALKER: I have reverted back the checks

AlphardWang (dismissed):

AnnService/src/Core/CommonDataStructure.cpp
```diff
@@ -59,9 +57,8 @@ ByteArray::operator= (const ByteArray& p_right)
     return *this;
 }
 
-
 ByteArray&
-ByteArray::operator= (ByteArray&& p_right)
+ByteArray::operator=(ByteArray&& p_right) noexcept
 {
     m_data = p_right.m_data;
```
asklar, line 63: shouldn't this move too?

PR #58, opened 2019-05-30
 Add new --append option to IndexBuilder to support adding new items to an existing index

MaJaHa95:
From what I can tell, there's no ability with the CLI to append data onto an existing index; running IndexBuilder starts afresh.

This PR adds an `--append` parameter to IndexBuilder, which attempts to load an existing index, and calls the `VectorIndex::AddIndex` function if it already exists.

I'm pretty new to C++ (from C#, so look out), so there are some changes I'm not totally confident about. But it seems pretty simple.

### DIFF
--- a/AnnService/src/IndexBuilder/main.cpp
+++ b/AnnService/src/IndexBuilder/main.cpp
@@ -55,7 +55,9 @@ int main(int argc, char* argv[])
         indexBuilder->SetParameter(iter.first.c_str(), iter.second.c_str());
     }
 
-    ErrorCode code;
+	std::shared_ptr<VectorSet> p_vectorSet = nullptr;
+	std::shared_ptr<MetadataSet> p_metaSet = nullptr;
+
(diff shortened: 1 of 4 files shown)


AnnService/src/IndexBuilder/main.cpp
```diff
@@ -87,9 +87,21 @@ int main(int argc, char* argv[])
             fprintf(stderr, "Failed to read input file.\n");
             exit(1);
         }
-        code = indexBuilder->BuildIndex(vectorReader->GetVectorSet(), vectorReader->GetMetadataSet());
-        indexBuilder->SaveIndex(options->m_outputFolder);
+
+        p_vectorSet = vectorReader->GetVectorSet();
```
MaJaHa95: Is there an issue with using the out-of-scope `auto vectorReader` here? I wasn't sure about the lifetime of it, whether there could be an issue. Again, new to all this.

PR #85, opened 2019-07-31
update fresh ANN implementation

MaggieQi:


AnnService/inc/Core/Common/Dataset.h
```diff
@@ -64,6 +65,7 @@ namespace SPTAG
                     else std::memset(data, -1, ((size_t)rows) * cols * sizeof(T));
                 }
             }
+            void SetName(const std::string name_) { name = name_; }
```
AlphardWang: please use const reference or move operator for such case

PR #93, opened 2019-08-12
fix some type definition in the Reader and add more support to create Reader

MaggieQi:


AnnService/inc/Core/Common/BKTree.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+ // Copyright (c) Microsoft Corporation. All rights reserved.
```
AlphardWang: space?

AnnService/src/Helper/VectorSetReaders/DefaultReader.cpp
```diff
@@ -244,7 +248,7 @@ DefaultReader::GetVectorSet() const
 
     std::ifstream inputStream;
     inputStream.open(m_vectorOutput, std::ifstream::binary);
-	inputStream.seekg(sizeof(uint32_t) + sizeof(uint32_t), std::ifstream::beg);
+	inputStream.seekg(sizeof(SizeType) + sizeof(DimensionType), std::ifstream::beg);
```
AlphardWang: please use space instead of tab

PR #105, opened 2019-10-08
fix parallel AddIndex issue, hash table size issue and configurable refine iteration issue

MaggieQi:


AnnService/inc/Core/Common/NeighborhoodGraph.h
```diff
@@ -25,7 +25,7 @@ namespace SPTAG
                                  m_iNeighborhoodSize(32),
                                  m_iNeighborhoodScale(2),
                                  m_iCEFScale(2),
-                                 m_iRefineIter(0),
+                                 m_iRefineIter(2),
```
AlphardWang: Maybe you can use macro to create init list.

AnnService/inc/Core/Common/WorkSpace.h
```diff
@@ -37,15 +37,15 @@ namespace SPTAG
...
+            int m_poolSize = 8191;
 
             // Could we use the second hash block.
             bool m_secondHash;
 
             // Record 2 hash tables.
             // [0~m_poolSize + 1) is the first block.
             // [m_poolSize + 1, 2*(m_poolSize + 1)) is the second block;
-            SizeType m_hashTable[(m_poolSize + 1) * 2];
+            std::unique_ptr<SizeType> m_hashTable;
```
AlphardWang: should be std::unique_ptr<SizeType[]>

AnnService/inc/Core/Common/WorkSpace.h
```diff
@@ -67,7 +67,9 @@ namespace SPTAG
 
             void Init(SizeType size)
             {
+                m_poolSize = 2 * size - 1;
```
AlphardWang: `m_poolSize` should be `power of 2 - 1`

AnnService/inc/Core/Common/WorkSpace.h
```diff
@@ -67,7 +67,9 @@ namespace SPTAG
 
             void Init(SizeType size)
             {
+                m_poolSize = (int)pow(2, log2(size) + 1) - 1;
```
AlphardWang: it should be at least (int)pow(2, (int)(log2(size) + 1)) - 1;
but I suggest to use simple loop here. a pow function may have precision issue.

AnnService/inc/Core/Common/BKTree.h
```diff
@@ -106,23 +107,36 @@ namespace SPTAG
...
+            inline SizeType sizePerTree() const {
+                std::shared_lock<std::shared_timed_mutex> lock(*m_lock);
+                return (SizeType)m_pTreeRoots.size() - m_pTreeStart.back(); 
+            }
 
             inline const std::unordered_map<SizeType, SizeType>& GetSampleMap() const { return m_pSampleCenterMap; }
 
+            void swap(BKTree& p_tree)
+            {
+                std::unique_lock<std::shared_timed_mutex> lock(*m_lock);
```
AlphardWang: should p_tree.m_lock be also locked?

AnnService/inc/Core/Common/Dataset.h
```diff
@@ -66,6 +66,8 @@ namespace SPTAG
                 }
             }
             void SetName(const std::string name_) { name = name_; }
+            std::string Name() const { return name; }
```
AlphardWang: return a const reference?

AnnService/inc/Core/Common/Dataset.h
```diff
@@ -200,6 +207,16 @@ namespace SPTAG
                 return true;
             }
 
+            bool Refine(const std::vector<SizeType>& indices, Dataset<T>& data)
+            {
+                SizeType R = (SizeType)(indices.size());
+                data.Initialize(R, cols);
+                for (SizeType i = 0; i < R; i++) {
+                    std::memcpy((void*)data.At(i), (void*)At(indices[i]), sizeof(T) * cols);
```
AlphardWang: what's the relationship between typename T and SizeType here?

AnnService/inc/Core/Common/KDTree.h
```diff
@@ -32,33 +33,45 @@ namespace SPTAG
...
             inline SizeType size() const { return (SizeType)m_pTreeRoots.size(); }
 
+            inline SizeType sizePerTree() const { 
+                std::shared_lock<std::shared_timed_mutex> lock(*m_lock);
+                return (SizeType)m_pTreeRoots.size() - m_pTreeStart.back(); 
+            }
+
+            void swap(KDTree& p_tree)
+            {
+                std::unique_lock<std::shared_timed_mutex> lock(*m_lock);
```
AlphardWang: same as BKT

AnnService/inc/Core/Common/Labelset.h
```diff
@@ -0,0 +1,111 @@
...
+                m_data.SetName("DeleteID");
+                m_lock.reset(new std::shared_timed_mutex);
+            }
+
+            void Initialize(SizeType capacity)
+            {
+                m_data.Initialize(capacity, 1);
+            }
+
+            inline size_t Size() const { return m_deleted.load(); }
```
AlphardWang: what does md_deleted mean here?

AnnService/inc/Core/Common/Labelset.h
```diff
@@ -0,0 +1,111 @@
...
+            {
+                return *m_data[key] == 1;
+            }
+
+            inline void Insert(const SizeType& key)
+            {
+                if (*m_data[key] != 1)
+                {
+                    std::shared_lock<std::shared_timed_mutex> lock(*m_lock);
+                    m_deleted++;
```
AlphardWang: should check if m_data[key] != 1 after locked

AnnService/inc/Core/Common/Dataset.h
MaggieQi: SizeType is used to represent the number of vectors. T is the vector data type.

AnnService/inc/Core/Common/BKTree.h
MaggieQi: m_lock is used to avoid doing search when trees are being updated. Here p_tree is the new version of trees which will not be used in any search process. It does not need to be locked.

AnnService/inc/Core/Common/Labelset.h
MaggieQi: It means how many items are deleted (label is 1).

AnnService/inc/Core/Common/Labelset.h
MaggieQi: This lock prevents any deletes when refining index. It does not prevent two same deletes at the same time.

AnnService/inc/Core/Common/Labelset.h
```diff
@@ -0,0 +1,102 @@
...
+            {
+                return *m_data[key] == 1;
+            }
+
+            inline void Insert(const SizeType& key)
+            {
+                if (*m_data[key] != 1)
+                {
+                    *m_data[key] = 1;
+                    m_deleted++;
```
AlphardWang: suggest to rename to m_inserted. BTW, please consider to use 1 bit instead of 1 byte to represent one occurrence.

Issue #104, opened 2019-09-18
An end to end example/tutorial for how to use SPTAG


scarlett2018:
It will be great help if we can provide an out of box example about how to use SPTAG in what kind of scenario.

SushantChatufale:
Looking forward to when this will be released!

PR #113, opened 2019-12-07
Add end-to-end walkthrough on usage of SPTAG

alyssaong1:
This provides an end-to-end example for developers to get started with SPTAG on an image dataset. 

Key additions:
- Jupyter notebook example that shows a user how to do visual similarity search on a public image dataset. Downloads images, extracts features from images, and using SPTAG to perform the similarity search. 
- Helper class (features_extractor.py) to do some image processing and extraction of features from images - this class is called from the notebook. 

This PR aims to address the following requests:
https://github.com/microsoft/SPTAG/issues/104


wilfoderek:
ANy practical example using this great tool?

johnt1838:
Yes it would be great if we had a practical example!

PR #131, opened 2020-03-05
Cheqi separate params

MaggieQi:
try to fix some bugs and separate some params


AnnService/inc/Core/Common/FineGrainedLock.h
```diff
@@ -16,43 +16,27 @@ namespace SPTAG
...
             const std::mutex& operator[](SizeType idx) const {
-                std::shared_lock<std::shared_timed_mutex> lock(*rwlock);
-                return *locks[idx];
+                unsigned index = hash_func((unsigned)idx);
+                return m_locks[index];
             }
         private:
-            std::unique_ptr<std::shared_timed_mutex> rwlock;
-            std::vector<std::shared_ptr<std::mutex>> locks;
+            static const int m_poolSize = 16383;
```
SkyInTheSea: It's abnormal to declare it as m_ as actually it's a static constant one which can not be modified, it's clearer as PoolSize

AnnService/inc/Core/Common/FineGrainedLock.h
```diff
@@ -16,43 +16,27 @@ namespace SPTAG
         class FineGrainedLock {
```
SkyInTheSea: I didn't check this class usage, but I think we do not allow a mutex object refer twice?
how can you resolve the issue that two threads refer the same mutex when index hash confliction?

AnnService/src/Core/BKT/BKTIndex.cpp
```diff
@@ -100,78 +100,151 @@ namespace SPTAG
...
+                p_space.m_iNumberOfCheckedLeaves++; \
+                p_space.m_NGQueue.insert(COMMON::HeapCell(nn_index, distance2leaf)); \
+            } \
+            if (p_space.m_NGQueue.Top().distance > p_space.m_SPTQueue.Top().distance) { \
+                m_pTrees.SearchTrees(this, p_query, p_space, m_iNumberOfOtherDynamicPivots + p_space.m_iNumberOfCheckedLeaves); \
+            } \
+        } \
+        p_query.SortResult(); \
+/*
+#define Search(CheckDeleted, CheckDuplicated) \
```
SkyInTheSea: suggest you not to define as this, it's hard to debug

AnnService/src/Core/VectorIndex.cpp
```diff
@@ -294,6 +294,42 @@ VectorIndex::DeleteIndex(ByteArray p_meta) {
 }
 
 
+ErrorCode
+VectorIndex::MergeIndex(const char* p_indexFilePath)
+{
+    std::string folderPath(p_indexFilePath);
+    if (!folderPath.empty() && *(folderPath.rbegin()) != FolderSep) folderPath += FolderSep;
```
SkyInTheSea: else ？

AnnService/src/Core/VectorIndex.cpp
```diff
@@ -294,6 +294,42 @@ VectorIndex::DeleteIndex(ByteArray p_meta) {
 }
 
 
+ErrorCode
+VectorIndex::MergeIndex(const char* p_indexFilePath)
```
SkyInTheSea: check nullptr firstly

AnnService/src/Core/VectorIndex.cpp
```diff
@@ -294,6 +294,42 @@ VectorIndex::DeleteIndex(ByteArray p_meta) {
...
+{
+    std::string folderPath(p_indexFilePath);
+    if (!folderPath.empty() && *(folderPath.rbegin()) != FolderSep) folderPath += FolderSep;
+    Helper::IniReader iniReader;
+    if (ErrorCode::Success != iniReader.LoadIniFile(folderPath + "indexloader.ini")) return ErrorCode::FailedOpenFile;
+
+    std::shared_ptr<VectorIndex> addIndex = CreateInstance(iniReader.GetParameter("Index", "IndexAlgoType", IndexAlgoType::Undefined),
+        iniReader.GetParameter("Index", "ValueType", VectorValueType::Undefined));
+    addIndex->LoadConfig(iniReader);
```
SkyInTheSea: we do not have return code for these functions? suppose not to succeed always

AnnService/src/Core/VectorIndex.cpp
```diff
@@ -294,6 +294,42 @@ VectorIndex::DeleteIndex(ByteArray p_meta) {
...
+    addIndex->LoadIndexData(folderPath);
+
+    std::shared_ptr<MetadataSet> pMetadata;
+    if (iniReader.DoesSectionExist("MetaData"))
```
SkyInTheSea: still, you need handle ELSE case

AnnService/src/Core/VectorIndex.cpp
```diff
@@ -294,6 +294,42 @@ VectorIndex::DeleteIndex(ByteArray p_meta) {
...
+        pMetadata.reset(new MemMetadataSet(folderPath + iniReader.GetParameter("MetaData", "MetaDataFilePath", std::string()),
+            folderPath + iniReader.GetParameter("MetaData", "MetaDataIndexPath", std::string())));
+    }
+
+#pragma omp parallel for schedule(dynamic,128)
+    for (SizeType i = 0; i < addIndex->GetNumSamples(); i++)
+        if (addIndex->ContainSample(i))
+        {
+            std::shared_ptr<MetadataSet> p_metaSet;
+            if (pMetadata != nullptr) {
```
SkyInTheSea: we do not need to check it every time, just place it before the loop once.

AnnService/src/Core/VectorIndex.cpp
```diff
@@ -294,6 +294,42 @@ VectorIndex::DeleteIndex(ByteArray p_meta) {
...
+                ByteArray meta = pMetadata->GetMetadata(i);
```
SkyInTheSea: can we ensure meta always valid here?

AnnService/src/Core/VectorIndex.cpp
```diff
@@ -294,6 +294,42 @@ VectorIndex::DeleteIndex(ByteArray p_meta) {
...
+                std::uint64_t offsets[2] = { 0, meta.Length() };
+                p_metaSet.reset(new MemMetadataSet(meta, ByteArray((std::uint8_t*)offsets, 2 * sizeof(std::uint64_t), false), 1));
```
SkyInTheSea: we always use sizeof(offsets) instead

AnnService/inc/Core/Common/FineGrainedLock.h
MaggieQi: If two threads try to refer the same mutex and try to lock it with lock_guard, one will wait until the other thread releases the lock.

AnnService/src/Core/VectorIndex.cpp
MaggieQi: not need for "else" since if the folderPath is empty,  iniReader.LoadIniFile will return ErrorCode::FailedOpenFile error.

AnnService/src/Core/VectorIndex.cpp
MaggieQi: if MetaData section does not exist, we do not need to load the pMetadata which will be nullptr.

AnnService/src/Core/VectorIndex.cpp
MaggieQi: no, we allow buildindex without metadata.

PR #132, opened 2020-03-20
Initial version of GPU Refinement code

bkarsin:
Does not achieve the same accuracy as CPU refinement.  Still looking into issue causing accuracy disparity.


bkarsin:
Closing pull request to create new request with updated GPU code that includes GPU refinement that achieves desired accuracy.

PR #136, opened 2020-04-01
Added AARCH64 support

mcelhennyi:
This pull request would allow for AARCH64 support. I made sure the ./test executable can run without error on an arm docker. 

The only thing I am unsure of is the use of __builtin_prefetch. SIMDe supports prefetch commands, but I am unsure how to map the enum over to SIMDe from the original use. Currently I have just used the __builtin_prefetch version in place of the __mm_prefetch.


pabs3:
@mcelhennyi I noticed that there are conflicts with the latest commit of the master branch, would you mind rebasing it? I'm considering adding it to the Debian package of SPTAG.

mcelhennyi:
@pabs3 I have actually stopped using SPTAG. So I won’t be able to rebase and test. Sorry about that.

pabs3:
Merging this issue will also fix #316 too.

pabs3:
I tried to create a new patch but SPTAG now uses AVX512 and SIMDe is missing three functions and is missing three native aliases (useful to avoid having to change lots of the code), the missing aliases is easily worked around though.

https://github.com/simd-everywhere/simde/issues/961
https://github.com/pabs3/SPTAG/compare/use-simd-everywhere

PR #138, opened 2020-05-14
memory estimations

MaggieQi:
add memory estimations

### DIFF
--- a/AnnService/inc/Core/BKT/Index.h
+++ b/AnnService/inc/Core/BKT/Index.h
@@ -96,6 +96,7 @@ namespace SPTAG
             ~Index() {}
 
             inline SizeType GetNumSamples() const { return m_pSamples.R(); }
+			inline SizeType GetNumDeletedSamples() const { return (SizeType)m_deletedID.Count(); }
             inline DimensionType GetFeatureDim() const { return m_pSamples.C(); }
         
             inline int GetCurrMaxCheck() const { return m_iMaxCheck; }
@@ -104,6 +105,7 @@ namespace SPTAG
             inline IndexAlgoType GetIndexAlgoType() const { return IndexAlgoType::BKT; }
             inline VectorValueType GetVectorValueType() const { return GetEnumValueType<T>(); }
             
+			inline bool IsIdenticalVector(const void* pX, const void* pY, float eps) const { return COMMON::DistanceUtils::ComputeL2Distance((const T*)pX, (const T*)pY, m_pSamples.C()) < eps; }
             inline float ComputeDistance(const void* pX, const void* pY) const { return m_fComputeDistance((const T*)pX, (const T*)pY, m_pSamples.C()); }
             inline const void* GetSample(const SizeType idx) const { return (void*)m_pSamples[idx]; }
             inline bool ContainSample(const SizeType idx) const { return !m_deletedID.Contains(idx); }
--- a/AnnService/inc/Core/BKT/Index.h
+++ b/AnnService/inc/Core/BKT/Index.h
@@ -96,7 +96,7 @@ namespace SPTAG
             ~Index() {}
 
             inline SizeType GetNumSamples() const { return m_pSamples.R(); }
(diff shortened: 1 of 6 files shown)


AnnService/inc/Core/BKT/Index.h
```diff
@@ -104,6 +105,7 @@ namespace SPTAG
             inline IndexAlgoType GetIndexAlgoType() const { return IndexAlgoType::BKT; }
             inline VectorValueType GetVectorValueType() const { return GetEnumValueType<T>(); }
             
+            inline bool IsIdenticalVector(const void* pX, const void* pY, float eps) const { return COMMON::DistanceUtils::ComputeL2Distance((const T*)pX, (const T*)pY, m_pSamples.C()) < eps; }
```
SkyInTheSea, line 108: will sptag only support l2? how about cosine distance?

AnnService/inc/Core/BKT/Index.h
```diff
@@ -96,6 +96,7 @@ namespace SPTAG
             ~Index() {}
 
             inline SizeType GetNumSamples() const { return m_pSamples.R(); }
+            inline SizeType GetNumDeletedSamples() const { return (SizeType)m_deletedID.Count(); }
```
SkyInTheSea, line 99: the function is a little confusing, it returns all deleted element count or samples?

AnnService/src/Core/VectorIndex.cpp
```diff
@@ -418,3 +432,41 @@ VectorIndex::LoadIndex(const std::string& p_config, const std::vector<ByteArray>
...
+    }
+    else {
+        return 0;
+    }
+    std::uint64_t unit = GetValueTypeSize(p_valuetype) * p_dimension + sizeof(std::uint64_t) + sizeof(SizeType) * p_neighborhoodSize + 1 + treeNodeSize * p_treeNumber;
+    return p_memory / unit;
+}
+
+
+std::uint64_t VectorIndex::EstimatedMemoryUsage(std::uint64_t p_vectorCount, DimensionType p_dimension, IndexAlgoType p_algo, VectorValueType p_valuetype, int p_treeNumber, int p_neighborhoodSize)
```
SkyInTheSea, line 454: could you do some experiment to give us some sense of the accuracy of this estimation? like give us a range the actual memory usage like [0,8, 1] of the estimated value.

what's more we need another non-static function to give the memory usage of the sptag instance

AnnService/inc/Core/BKT/Index.h
MaggieQi: cosine is also supported. It can be more accurate using L2 to judge whether two vectors are the same.

AnnService/inc/Core/BKT/Index.h
MaggieQi: It returns all deleted element count.

AnnService/src/Core/VectorIndex.cpp
MaggieQi: This calculation only skip some sizeof(int) and dynamic memory usage (it related to the thread number).

We have non-static function to give memory usage. I think I have added in the wrapper.

AnnService/inc/Core/BKT/Index.h
SkyInTheSea: the meaning of the function is “identical" upon specified space instead of exact match, like for cosine distance, we store normalized vector, but the input vector may be unnormalized, so we should pick corresponding distance function.

AnnService/inc/Core/BKT/Index.h
SkyInTheSea: then could you change the function name?

AnnService/inc/Core/BKT/Index.h
MaggieQi: ok, let me change it.

AnnService/inc/Core/BKT/Index.h
MaggieQi: ok

PR #141, opened 2020-05-23
fix delete error code

MaggieQi:


AnnService/src/Core/BKT/BKTIndex.cpp
```diff
@@ -448,8 +448,8 @@ namespace SPTAG
         template <typename T>
         ErrorCode Index<T>::DeleteIndex(const SizeType& p_id) {
             std::shared_lock<std::shared_timed_mutex> sharedlock(m_dataDeleteLock);
```
SkyInTheSea, line 450: it should be a exclusive lock, otherwise you can consider the case two concurrent delete thread with same ID can both succeed and m_inserted will be added twice

PR #142, opened 2020-06-04
add errorcodes and fix refine modify index issue

MaggieQi:


AnnService/src/Core/KDT/KDTIndex.cpp
```diff
@@ -381,6 +385,8 @@ namespace SPTAG
         template <typename T>
         ErrorCode Index<T>::AddIndex(const void* p_data, SizeType p_vectorNum, DimensionType p_dimension, std::shared_ptr<MetadataSet> p_metadataSet, bool p_withMetaIndex)
         {
+            if (p_data == nullptr || p_vectorNum == 0 || p_dimension == 0) return ErrorCode::EmptyData;
```
SkyInTheSea, line 388: for add index scenario, we may also consider inconsistent dimension as mismatch errror

MaggieQi: We have checked the data dimension mismatch in the next few lines.

PR #145, opened 2020-06-18
Cheqi add interfaces

MaggieQi:
*add mergeindex from indexblobs
*add search interface to support batch mode
*fix metamapping load issue

### DIFF
--- a/AnnService/src/Core/VectorIndex.cpp
+++ b/AnnService/src/Core/VectorIndex.cpp
@@ -114,8 +114,10 @@ VectorIndex::BuildMetaMapping()
 {
     m_pMetaToVec.reset(new std::unordered_map<std::string, SizeType>);
     for (SizeType i = 0; i < m_pMetadata->Count(); i++) {
-        ByteArray meta = m_pMetadata->GetMetadata(i);
-        m_pMetaToVec->emplace(std::string((char*)meta.Data(), meta.Length()), i);
+        if (ContainSample(i)) {
+            ByteArray meta = m_pMetadata->GetMetadata(i);
+            m_pMetaToVec->emplace(std::string((char*)meta.Data(), meta.Length()), i);
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


AnnService/src/Core/VectorIndex.cpp
```diff
@@ -114,8 +114,10 @@ VectorIndex::BuildMetaMapping()
 {
     m_pMetaToVec.reset(new std::unordered_map<std::string, SizeType>);
     for (SizeType i = 0; i < m_pMetadata->Count(); i++) {
-        ByteArray meta = m_pMetadata->GetMetadata(i);
-        m_pMetaToVec->emplace(std::string((char*)meta.Data(), meta.Length()), i);
+        if (ContainSample(i)) {
+            ByteArray meta = m_pMetadata->GetMetadata(i);
+            m_pMetaToVec->emplace(std::string((char*)meta.Data(), meta.Length()), i);
```
SkyInTheSea, line 119: if i not contains in m_pMetadata expected? should we add some error handle code?

AnnService/src/Core/VectorIndex.cpp
```diff
@@ -269,9 +276,13 @@ VectorIndex::BuildIndex(std::shared_ptr<VectorSet> p_vectorSet,
 
 
 ErrorCode
-VectorIndex::SearchIndex(const void* p_vector, int p_neighborCount, bool p_withMeta, BasicResult* p_results) const {
-    QueryResult res(p_vector, p_neighborCount, p_withMeta, p_results);
-    SearchIndex(res);
+VectorIndex::SearchIndex(const void* p_vector, int p_vectorCount, int p_neighborCount, bool p_withMeta, BasicResult* p_results) const {
+    size_t vectorSize = GetValueTypeSize(GetVectorValueType()) * GetFeatureDim();
+#pragma omp parallel for schedule(dynamic,10)
+    for (int i = 0; i < p_vectorCount; i++) {
```
SkyInTheSea, line 282: searchindex need follow the policy to leverage only one thread, even for batch search(though fresh ANN index didn't support batch search), as we use request thread pool to control processor usage, every request thread consume several core is not our expectation

AnnService/src/Core/VectorIndex.cpp
MaggieQi: Here we check whether this vector has been deleted or not. If we do not check this, the metadata will be overwritten by deleted items.

AnnService/src/Core/VectorIndex.cpp
MaggieQi: This interface is only used for parallel search in the python and java wrappers to speedup the offline search.

PR #155, opened 2020-07-22
Fix linux build

MaggieQi:


AnnService/src/Core/VectorIndex.cpp
```diff
@@ -373,7 +373,7 @@ VectorIndex::MergeIndex(VectorIndex* p_addindex, int p_threadnum, IAbortOperatio
             if (p_abort != nullptr && p_abort->ShouldAbort()) 
             {
                 ret = ErrorCode::ExternalAbort;
-                break;
+                continue;
```
SkyInTheSea, line 376: why use continue instead of break; continue will roll the total loop then quit

PR #167, opened 2020-08-28
Fix simd instructions

MaggieQi:


DovahYol (approved):
Reviewed. 
I've verified that it can be compiled as a CMake project in a machine that doesn't support AVX2. 
This is my basic settings:
OS: WSL (ubuntu 16.04 LTS)
CMAKE_BUILD_TYPE: Release/Debug

PR #171, opened 2020-09-03
fix save graph size

MaggieQi:


AnnService/inc/Core/Common/NeighborhoodGraph.h
```diff
@@ -478,12 +478,21 @@ namespace SPTAG
...
+                auto ptr = f_createIO();
+                if (ptr == nullptr || !ptr->Initialize(sGraphFilename.c_str(), std::ios::binary | std::ios::out)) return ErrorCode::FailedCreateFile;
+                return SaveGraph(ptr);
             }
 
             ErrorCode SaveGraph(std::shared_ptr<Helper::DiskPriorityIO> output) const
             {
-                return m_pNeighborhoodGraph.Save(output);
+                IOBINARY(output, WriteBinary, sizeof(SizeType), (char*)&m_iGraphSize);
+                IOBINARY(output, WriteBinary, sizeof(DimensionType), (char*)&m_iNeighborhoodSize);
```
SkyInTheSea, line 490: every IO can fail, do you have return code for IOBinary?

SkyInTheSea (approved):

AnnService/inc/Core/Common/NeighborhoodGraph.h
MaggieQi: yes, IOBINARY will checkout the read and write bytes and return ErrorCode when the io fails.

PR #188, opened 2021-02-03
fix add perf and linux compiling issues

MaggieQi:
Capture

### DIFF
--- a/AnnService/inc/Core/Common/Dataset.h
+++ b/AnnService/inc/Core/Common/Dataset.h
@@ -19,6 +19,7 @@ namespace SPTAG
             T* data = nullptr;
             bool ownData = false;
             SizeType incRows = 0;
+            SizeType maxRows;
             SizeType rowsInBlock;
             SizeType rowsInBlockEx;
             std::vector<T*> incBlocks;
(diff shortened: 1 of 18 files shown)


AnnService/inc/Core/Common/Dataset.h
```diff
@@ -47,8 +49,10 @@ namespace SPTAG
                     if (data_ != nullptr) memcpy(data, data_, ((size_t)rows) * cols * sizeof(T));
                     else std::memset(data, -1, ((size_t)rows) * cols * sizeof(T));
                 }
-                rowsInBlock = rowsInBlock_;
-                incBlocks.reserve((static_cast<std::int64_t>(capacity_) + rowsInBlock - 1) / rowsInBlock);
+                maxRows = capacity_;
+                rowsInBlockEx = static_cast<SizeType>(ceil(log2(rowsInBlock_)));
+                rowsInBlock = (1 << rowsInBlockEx) - 1;
+                incBlocks.reserve((static_cast<std::int64_t>(capacity_) + rowsInBlock) >> rowsInBlockEx);
```
SkyInTheSea, line 55: you may need check some edge case like rowInBlock_ == 1, then looks rowsInBlockEx = 0, then rowsInBlock = 0 also, which looks not right

SkyInTheSea (approved):

AnnService/inc/Core/Common/Dataset.h
MaggieQi: ok, you are right, let me fix it

PR #225, opened 2021-08-21
Fix typos in Quick Start doc

blueviggen:
The command line image names are server, client, aggregator, indexbuilder and indexsearcher (in all lowercase).


pabs3:
This is missing a typo fix: `indexbuiler` should be `indexbuilder`.

PR #240, opened 2021-10-14
Add PQ support for memory and SSD SPTAG

PhilipBAdams:


PhilipBAdams:
Need to make change to write quantizer info on index save

AnnService/inc/Core/Common/BKTree.h
```diff
@@ -255,7 +262,9 @@ namespace SPTAG
                 }
                 args.ClearCounts();
                 args.ClearDists(MaxDist);
-                currDist = KmeansAssign(data, indices, first, batchEnd, args, false, 0);
```
MaggieQi, line 258: why this line need to change?

PhilipBAdams: I don't think it does. It looks like I added it last summer, but I do not remember why. I will run a recall test without the change and check for sure.

- [ ]  Recall test w/o KmeansAssign baseSquare
- [ ] Check for other unnecessary baseSquare changes

PhilipBAdams: Recall matches without the change. I think it was added by mistake with the other normalization changes. Should be fixed in latest version

MaggieQi (changes requested):

AnnService/inc/SSDServing/VectorSearch/SearchSsdIndex.h
```diff
@@ -518,7 +525,12 @@ namespace SPTAG {
                 LOG(Helper::LogLevel::LL_Info, "\nFinish ANN Search...\n");
 
                 std::shared_ptr<SPTAG::VectorSet> vectorSet;
-                if (!COMMON_OPTS.m_vectorPath.empty() && fileexists(COMMON_OPTS.m_vectorPath.c_str())) {
+                if (!SPTAG::COMMON::DistanceUtils::Quantizer && p_opts.m_rerankFilePath.empty())
+                {
+                    p_opts.m_rerankFilePath = COMMON_OPTS.m_vectorPath;
```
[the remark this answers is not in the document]
PhilipBAdams: If we want to rerank quantized vectors with original full-dim ones, we need to have different rerank path from vector path. I set them to be the same if we do not have value for rerankFile

AnnService/inc/SSDServing/SelectHead_BKT/ParameterDefinitionList.h
```diff
@@ -29,4 +29,7 @@ DefineSelectHeadParameter(m_recursiveCheckSmallCluster, bool, true, "RecursiveCh
 DefineSelectHeadParameter(m_printSizeCount, bool, true, "PrintSizeCount")
 DefineSelectHeadParameter(m_selectType, std::string, "BKT", "SelectHeadType")
 
+DefineSelectHeadParameter(m_datasetRowsInBlock, int, 1024 * 1024, "DataBlockSize")
```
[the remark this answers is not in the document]
PhilipBAdams: It is feature from internal version for large datasets. I think it is better to match internal/external versions for less duplicated effort, but I can remove from the PR if you would like.

AnnService/src/IndexSearcher/main.cpp
```diff
@@ -63,6 +65,10 @@ class SearcherOptions : public Helper::ReaderOptions
     int m_genTruth = 0;
 
     int m_debugQuery = -1;
+
+    std::string m_quantizerFile;
```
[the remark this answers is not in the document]
PhilipBAdams: QuantizerFile as a setting can be removed, it is from before I add the load with index. I think ADC should be option here though, since it is a search property not an index property

AnnService/src/Core/BKT/BKTIndex.cpp
```diff
@@ -196,7 +196,7 @@ namespace SPTAG
                 SizeType nn_index = node[i]; \
                 if (nn_index < 0) break; \
                 if (p_space.CheckAndSet(nn_index)) continue; \
-                float distance2leaf = m_fComputeDistance(p_query.GetTarget(), (m_pSamples)[nn_index], GetFeatureDim()); \
+                float distance2leaf = m_fComputeDistance(p_query.GetQuantizedTarget(), (m_pSamples)[nn_index], GetFeatureDim()); \
```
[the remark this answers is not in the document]
PhilipBAdams: I think it is cleaner to describe it this way, because when we have no quantizer set, we can think of it as the 'identity quantizer', that quantizes every vector to itself. Also, it is easier to make default option (GetTarget) most compatible (reconstructed vector), and explicitly switch to GetQuantizedTarget when we only need distance calculation ability. 

If you have strong opinion on it, I am willing to change it though

AnnService/src/SSDServing/IndexBuildManager/main.cpp
```diff
@@ -187,7 +205,7 @@ namespace SPTAG {
 				}
 				auto vectorSet = vectorReader->GetVectorSet();
 				auto querySet = queryReader->GetVectorSet();
-				if (COMMON_OPTS.m_distCalcMethod == DistCalcMethod::Cosine) vectorSet->Normalize(search_ssd_opts.m_iNumberOfThreads);
+				if (COMMON_OPTS.m_distCalcMethod == DistCalcMethod::Cosine && !SPTAG::COMMON::DistanceUtils::Quantizer) vectorSet->Normalize(search_ssd_opts.m_iNumberOfThreads);
```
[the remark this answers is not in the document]
PhilipBAdams: Yeah, we can, it will be cleaner. Thanks :)

docs/GettingStart.md
```diff
@@ -40,6 +42,9 @@
   -dft, --dataFileType <value>  original data file type. (TXT, or DEFAULT)
   -b, --batchsize <value>       Batch query size.
   -g, --gentruth <value>        Generate truth file.
+  -q, --debugquery <value>      Debug query number.
+  -pq, --quantizer <value>      Quantizer File
```
MaggieQi, line 46: remove this line?

AnnService/inc/Core/Common/BKTree.h
```diff
@@ -149,8 +149,15 @@ namespace SPTAG
                     for (DimensionType j = 0; j < args._D; j++) currCenters[j] /= args.counts[k];
 
                     if (args._M == DistCalcMethod::Cosine) {
-                        COMMON::Utils::Normalize(currCenters, args._D, COMMON::Utils::GetBase<T>());
+                        if (SPTAG::COMMON::DistanceUtils::Quantizer) {
```
MaggieQi, line 152: If we move the check into GetBase, then this line can also keep original.

MaggieQi (changes requested):

AnnService/inc/SSDServing/VectorSearch/SearchSsdIndex.h
```diff
@@ -73,7 +75,7 @@ namespace SPTAG {
                             }
                             else if (vectorSet != nullptr) {
                                 float dist = results[i].GetResult(j)->Dist;
-                                float truthDist = index->ComputeDistance(querySet->GetVector(i), vectorSet->GetVector(id));
```
MaggieQi, line 76: Why index->ComputeDistance cannot return correct distance?

MaggieQi: If it cannot, could you try to use SPTAG::COMMON::DistanceUtils::ComputeDistance<T>(const T* p1, const T* p2, DimensionType length, SPTAG::DistCalcMethod distCalcMethod) interface?

AnnService/inc/SSDServing/VectorSearch/SearchSsdIndex.h
```diff
@@ -531,13 +537,15 @@ namespace SPTAG {
 
                 if (p_opts.m_rerank > 0 && vectorSet != nullptr) {
                     LOG(Helper::LogLevel::LL_Info, "\n Begin rerank...\n");
+                    COMMON::DistanceUtils::Quantizer.reset();
+                    auto distCalc = SPTAG::COMMON::DistanceCalcSelector<ValueType>(COMMON_OPTS.m_distCalcMethod);
```
MaggieQi, line 541: This part has not changed? It can be replaced by the COMMON::DistanceUtils::ComputeDistance().

AnnService/src/IndexSearcher/main.cpp
```diff
@@ -114,7 +118,8 @@ float CalcRecall(VectorIndex* index, std::vector<QueryResult>& results, const st
             for (SizeType id : truth[i]) {
                 float truthDist = 0.0;
                 if (vectorSet != nullptr) {
-                    truthDist = index->ComputeDistance(querySet->GetVector(i), vectorSet->GetVector(id));
+                    auto distCalc = COMMON::DistanceCalcSelector<T>(index->GetDistCalcMethod());
+                    truthDist = distCalc((const T*) querySet->GetVector(i), (const T*)vectorSet->GetVector(id), querySet->Dimension());
```
MaggieQi, line 122: replace by COMMON::DistanceUtils::ComputeDistance<T>()？

docs/GettingStart.md
```diff
@@ -135,6 +141,8 @@ ResultNum=10
 MaxCheck=2048
 MaxDistRatio=8.0
 SearchPostingPageLimit=3
+rerank=10
+rerankFilePath=base.1B.u8bin
```
MaggieQi, line 145: remove rerankFilePath?

MaggieQi (changes requested):

MaggieQi (approved):

MaggieQi (approved):

PR #242, opened 2021-10-27
fix balanced data partition limit overflow

MaggieQi:


AnnService/src/Core/VectorIndex.cpp
```diff
@@ -721,36 +721,9 @@ void VectorIndex::ApproximateRNG(std::shared_ptr<VectorSet>& fullVectors, std::u
...
-        }
-
-        return a.node < b.node;
-    };
-} g_edgeComparer;
-
 void VectorIndex::SortSelections(std::vector<Edge>* selections) {
-  std::sort(selections->begin(), selections->end(), g_edgeComparer);
+    EdgeCompare edgeComparer;
+    std::sort(selections->begin(), selections->end(), edgeComparer);
```
SkyInTheSea, line 726: why use local one instead of global

MaggieQi: it will cause compiling issue.

PR #253, opened 2021-12-21
Add script to install Linux dependencies and install SPTAG

christiancosgrove:


christiancosgrove:
Didn't see the `Dockerfile` installation instructions, so this script probably isn't necessary.

PR #260, opened 2022-01-28
Quantizer EXE, OPQ Support

PhilipBAdams:
* Add PQ Quantizer Training support

* Add OPQ support for index build/index search


AnnService/src/Core/VectorIndex.cpp
```diff
@@ -368,6 +368,9 @@ VectorIndex::SaveIndexToFile(const std::string& p_file, IAbortOperation* p_abort
     IOBINARY(fp, WriteBinary, sizeof(configSize), (char*)&configSize);
     if ((ret = SaveIndexConfig(fp)) != ErrorCode::Success) return ret;
     configSize = fp->TellP() - sizeof(configSize);
+    if (ErrorCode::Success == ret && m_pQuantizer) {
```
MaggieQi, line 371: Why we need to change the load quantizer position? If you want to change, then you need to change all of the save index and load index as well as the buffersizes and filenames order you get from the index.

AnnService/src/IndexBuilder/main.cpp
```diff
@@ -101,9 +105,14 @@ int main(int argc, char* argv[])
         code = indexBuilder->BuildIndex(vectorReader->GetVectorSet(), vectorReader->GetMetadataSet(), options->m_metaMapping, options->m_normalized);
     }
     else {
+        indexBuilder->SetQuantizerFileName(options->m_quantizerFile.substr(options->m_quantizerFile.find_last_of("/\\") + 1));
         code = indexBuilder->BuildIndex(options->m_normalized);    
     }
-    if (code == ErrorCode::Success) indexBuilder->SaveIndex(options->m_outputFolder);
+    if (code == ErrorCode::Success)
+    {
+
```
MaggieQi, line 113: remove the empty line?

MaggieQi (changes requested):

AnnService/inc/Core/SPANN/ExtraFullGraphSearcher.h
```diff
@@ -106,9 +106,9 @@ namespace SPTAG
                     auto curIndexFile = f_createAsyncIO();
                     if (curIndexFile == nullptr || !curIndexFile->Initialize(curFile.c_str(), std::ios::binary | std::ios::in, 
 #ifdef BATCH_READ
-                        p_opt.m_searchInternalResultNum, 2, 2, p_opt.m_iSSDNumberOfThreads
+                        max(p_opt.m_searchInternalResultNum*m_vectorInfoSize, 1 << 12), 2, 2, p_opt.m_iSSDNumberOfThreads
```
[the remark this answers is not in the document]
PhilipBAdams: Sorry, can you share the PR? I had some trouble finding it

AnnService/src/Core/VectorIndex.cpp
PhilipBAdams: Before the local-to-index quantizer change, it was needed to load the quantizer first so the right distance function would be picked. Now, it is not needed - I think it is changed in all the right places, but I can revert it.

MaggieQi: It is better to keep original load order : )

PhilipBAdams: returned the load order in most recent commit

MaggieQi (approved):

PR #269, opened 2022-03-07
make sure deletedID block is initialized

PhilipBAdams:
We encounter case where `nullptr` is accessed during checking `m_deletedID`. This change makes sure that deletedID buffer is allocated at initialization time to prevent null reference.


AnnService/inc/Core/Common/Labelset.h
```diff
@@ -26,7 +26,7 @@ namespace SPTAG
 
             void Initialize(SizeType size, SizeType blockSize, SizeType capacity)
             {
-                m_data.Initialize(size, 1, blockSize, capacity);
+                m_data.Initialize(size, 1, blockSize, capacity, nullptr, false);
```
MaggieQi, line 29: this change is not necessary.

MaggieQi (changes requested):

MaggieQi (approved):

PR #272, opened 2022-03-16
Fix float quantizer

MaggieQi:


AnnService/inc/Core/Common/BKTree.h
```diff
@@ -164,14 +170,21 @@ namespace SPTAG
...
-                    float* currCenters = args.newCenters + k * args._D;
-                    for (DimensionType j = 0; j < args._D; j++) currCenters[j] /= args.counts[k];
-
+                    float* currCenters = args.newCenters + k * args._RD;
+                    for (DimensionType j = 0; j < args._RD; j++) {
+                        currCenters[j] /= args.counts[k];
+                    }
                     if (args._M == DistCalcMethod::Cosine) {
-                        COMMON::Utils::Normalize(currCenters, args._D, COMMON::Utils::GetBase<T>());
+                        COMMON::Utils::Normalize(currCenters, args._RD, COMMON::Utils::GetBase<T>());
```
MaggieQi, line 178: I am not sure whether Cosine still need to normalize. Could you help me to double check it?

PhilipBAdams: I think there may be risk of too much distortion when quantizing the centers if we normalize them first. I will run a test with both configurations to check if there is regression.

PR #279, opened 2022-04-23
Support cleaning up batch requests (windows)

PhilipBAdams:


AnnService/src/Helper/AsyncFileReader.cpp
```diff
@@ -82,39 +82,48 @@ namespace SPTAG {
...
+
+        void BatchReadFileAsync(std::vector<std::shared_ptr<Helper::DiskPriorityIO>>& handlers, AsyncReadRequest* readRequests, int num)
+        {
+            
+            CallOnAppropriateBatch(handlers, readRequests, num, &DiskPriorityIO::BatchReadFile);
 
             for (int i = 0; i < num; i++) {
                 AsyncReadRequest* readRequest = &(readRequests[i]);
                 if (readRequest->m_readSize == 0) continue;
-                readRequest->m_readSize = 0;
```
MaggieQi, line 112: We need to clean readSize here since we use readSize to judge whether we need to issue a read request or not.

AnnService/src/Helper/AsyncFileReader.cpp
```diff
@@ -82,39 +82,48 @@ namespace SPTAG {
...
-                if (readRequest->m_success)
+                
+                if (readRequest->m_success && readRequest->m_callback)
```
MaggieQi, line 119: If we use the readSize to judge whether the request is available then we do not need to test the m_callback? Or we change the logic to use m_callback to test the availability?

AnnService/src/Core/SPANN/SPANNIndex.cpp
```diff
@@ -212,7 +212,7 @@ namespace SPTAG
                 workSpace->m_postingIDs.clear();
 
                 float limitDist = p_queryResults->GetResult(0)->Dist * m_options.m_maxDistRatio;
-                for (int i = 0; i < m_options.m_searchInternalResultNum; ++i)
+                for (int i = 0; i < min(m_options.m_searchInternalResultNum, p_queryResults->GetResultNum()); ++i)
```
MaggieQi, line 215: We may need to enlarge the queryset before the headindex search? Otherwise we cannot increase the internalresultnum to get better recall.

AnnService/src/Helper/AsyncFileReader.cpp
PhilipBAdams: With the cleanup, there needs to be an extra state where the read has been process, but the resource has not been cleaned up.

In this implementation, we use the m_callback to check if it has been processed, and then the readSize to judge if the resource has been cleaned up

AnnService/src/Core/SPANN/SPANNIndex.cpp
PhilipBAdams: We will segfault if p_queryResults is smaller than searchInternalResultNum

Maybe it is better to just iterate over p_queryResults->GetResultNum()?

AnnService/src/Helper/AsyncFileReader.cpp
MaggieQi: Then we need to do the line 112 after the cleanup to make sure that the readSize is cleaned after the request is executed since when a posting contains 0 items, it will be skipped in the ExtraSearcher without setting the readSize to 0.

AnnService/src/Core/SPANN/SPANNIndex.cpp
MaggieQi: no, it should be searchInternalResultNum. We need to ensure the GetResultNum() is equal or larger than searchInternResultNum.

AnnService/src/Helper/AsyncFileReader.cpp
PhilipBAdams: I move the skip before the m_extraSearcher->SearchIndex call - it handles this case, and also means we will not sacrifice a read if one of the postings contains 0 items but we have more than `searchInternalResultNum` candidates from the head

AnnService/src/Core/SPANN/SPANNIndex.cpp
PhilipBAdams: I agree that callers should provide `QueryResultSet`s with more than `searchInternalResultNum` results, which we enforce in our `SSDServing`. 

However, if we get to this point with the result set being too small, I think it is better to complete the search than to segfault. I can add a debug assert if you would like to warn the caller about it

AnnService/inc/Core/BKT/Index.h
```diff
@@ -122,8 +122,8 @@ namespace SPTAG
                 return 1.0f - xy / (sqrt(xx) * sqrt(yy));
             }
             inline float ComputeDistance(const void* pX, const void* pY) const { return m_fComputeDistance((const T*)pX, (const T*)pY, m_pSamples.C()); }
-            inline const void* GetSample(const SizeType idx) const { return (void*)m_pSamples[idx]; }
-            inline bool ContainSample(const SizeType idx) const { return idx < m_pSamples.R() && !m_deletedID.Contains(idx); }
+            inline const void* GetSample(const SizeType idx) const { return (idx >= 0 && idx < m_pSamples.R()) ? (void*)m_pSamples[idx] : nullptr; }
```
MaggieQi, line 125: Philip, it is better not to do the check in the GetSample since it will be called many times in the search. Add the check in the ContainSample is ok. So if the wrapper want to get sample, check the contain sample first.

AnnService/inc/Core/KDT/Index.h
```diff
@@ -120,10 +120,11 @@ namespace SPTAG
                 return 1.0f - xy / (sqrt(xx) * sqrt(yy));
             }
             inline float ComputeDistance(const void* pX, const void* pY) const { return m_fComputeDistance((const T*)pX, (const T*)pY, m_pSamples.C()); }
-            inline const void* GetSample(const SizeType idx) const { return (void*)m_pSamples[idx]; }
-            inline bool ContainSample(const SizeType idx) const { return idx < m_pSamples.R() && !m_deletedID.Contains(idx); }
+            inline const void* GetSample(const SizeType idx) const { return (idx >= 0 && idx < m_pSamples.R()) ? (void*)m_pSamples[idx] : nullptr; }
```
MaggieQi, line 123: The same as above.

AnnService/src/Core/SPANN/SPANNIndex.cpp
```diff
@@ -212,18 +212,18 @@ namespace SPTAG
...
                 {
                     auto res = p_queryResults->GetResult(i);
+                    auto postingID = res->VID;
+
+                    // Overwrite the posting ID with the actual vector ID
                     if (res->VID == -1) break;
                     res->VID = static_cast<SizeType>((m_vectorTranslateMap.get())[res->VID]);
+
+                    // Don't do disk reads for irrelevant pages
+                    if (workSpace->m_postingIDs.size() >= m_options.m_searchInternalResultNum || (limitDist > 0.1 && res->Dist > limitDist) || !m_extraSearcher->CheckValidPosting(postingID)) continue;
```
MaggieQi, line 225: Let me try to reorganize the code a little bit.

MaggieQi (approved):

PR #283, opened 2022-04-30
Reduce memory spann

MaggieQi:


PhilipBAdams (dismissed):

Wrappers/src/CoreInterface.cpp
```diff
@@ -137,12 +137,27 @@ AnnIndex::SetSearchParam(const char* p_name, const char* p_value, const char* p_
...
+
+void
+AnnIndex::SetQuantizerADC(bool p_adc)
+{
+    if (nullptr != m_index) return m_index->SetQuantizerADC(p_adc);
+}
+
+
 std::shared_ptr<QueryResult>
 AnnIndex::Search(ByteArray p_data, int p_resultNum)
```
PhilipBAdams, line 156: I need to change the internal search logic to handle this case - currently, we require the QueryResult to have the m_pQuantizedVector set when SearchIndex is called, to avoid risk of calling QuantizeVector multiple times.

We can do it all internally instead with a bit more complexity - I can make the change separately

MaggieQi: Hi Philip, I tested the python wrapper with quantization in Tutorial.ipynb. But it seems the recall of PQ quantizer has some issue. Could you help me to double check the Tutorial.ipynb?

PhilipBAdams (dismissed):

PhilipBAdams (approved):

PR #297, opened 2022-05-31
support data compression & delta-encoding of posting list

suiguoxin:


AnnService/inc/Core/SPANN/ExtraFullGraphSearcher.h
```diff
@@ -581,8 +778,8 @@ namespace SPTAG
                     }
 
                     m_listInfos[i].listOffset = (static_cast<uint64_t>(m_listPageOffset + pageNum) << PageSizeEx);
-                    m_listInfos[i].listEleCount = min(m_listInfos[i].listEleCount, (min(static_cast<int>(m_listInfos[i].listPageCount), p_postingPageLimit) << PageSizeEx) / m_vectorInfoSize);
-                    m_listInfos[i].listPageCount = static_cast<std::uint16_t>(ceil((m_vectorInfoSize * m_listInfos[i].listEleCount + m_listInfos[i].pageOffset) * 1.0 / (1 << PageSizeEx)));
+                    // m_listInfos[i].listEleCount = min(m_listInfos[i].listEleCount, (min(static_cast<int>(m_listInfos[i].listPageCount), p_postingPageLimit) << PageSizeEx) / m_vectorInfoSize);
```
PhilipBAdams, line 781: If it is not needed anymore, we can remove it instead of comment out

AnnService/inc/Core/SPANN/ExtraFullGraphSearcher.h
```diff
@@ -602,14 +799,51 @@ namespace SPTAG
                     }
                 }
 
+                if (ptr->ReadBinary(sizeof(m_enableDeltaEncoding), reinterpret_cast<char *>(&m_enableDeltaEncoding)) != sizeof(m_enableDeltaEncoding))
```
PhilipBAdams, line 802: Does it mean we will not be able to load SSD index built by old version of SPTAG, since the head info will not contain these config bytes? 

I don't think we can break compatibility like this

suiguoxin: > Does it mean we will not be able to load SSD index built by old version of SPTAG, since the head info will not contain these config bytes?
> 
> I don't think we can break compatibility like this

Another approach is to add these parameters in `m_extraSearcher->SearchIndex` function, with which we may have to change the Nuget package interface. Which approach do we prefer?

AnnService/inc/Core/SPANN/Compressor.h
```diff
@@ -0,0 +1,203 @@
...
+                {
+                    LOG(Helper::LogLevel::LL_Error, "ZSTD_createDCtx() failed! \n");
+                    exit(1);
+                }
+                size_t const decomp_size = ZSTD_decompress_usingDDict(dctx,
+                                                                      (void *)decomp_buffer.data(), est_decomp_size, src, srcSize, ddict);
+                if (ZSTD_isError(decomp_size))
+                {
+                    LOG(Helper::LogLevel::LL_Error, "ZSTD decompress error %s, \n", ZSTD_getErrorName(decomp_size));
+                    exit(1);
```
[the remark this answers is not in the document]
suiguoxin: `exit(1)` is largely used in `LoadingHeadInfo`  [here](https://github.com/microsoft/SPTAG/blob/main/AnnService/inc/Core/SPANN/ExtraFullGraphSearcher.h#L519), so you mean we just cannot use `exit(1)` in `SearchIndex` ?

AnnService/inc/Core/SPANN/Compressor.h
```diff
@@ -0,0 +1,203 @@
...
+
+                return dictSize;
+            }
+
+            std::string GetDictBuffer()
+            {
+                return dictBuffer;
+            }
+
+            void SetDictBuffer(std::string buffer)
```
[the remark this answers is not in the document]
suiguoxin: > Should it take an rvalue reference so we can move the value?

Fixed, thanks

AnnService/inc/Core/SPANN/ExtraFullGraphSearcher.h
suiguoxin: > If it is not needed anymore, we can remove it instead of comment out

I'm not sure if these two lines are useful under no-compression setting, @MaggieQi can we just remove these two lines?

AnnService/inc/Core/SPANN/Compressor.h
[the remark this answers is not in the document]
PhilipBAdams: Ideally, we would not be using it in LoadingHeadInfo either. But it is most important to remove it from SearchIndex

AnnService/inc/Core/SPANN/ExtraFullGraphSearcher.h
suiguoxin: Added a codebranch to keep these 2 lines for no-compression setting.

AnnService/inc/Core/SPANN/Compressor.h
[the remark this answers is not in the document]
suiguoxin: Fixed with exception try/catch

AnnService/inc/Core/SPANN/Compressor.h
```diff
@@ -0,0 +1,203 @@
...
+                dictBuffer = buffer;
+                CreateDDict();
+            }
+
+            std::string Compress(const std::string &src, const bool useDict)
+            {
+                return useDict ? CompressWithDict(src) : CompressWithoutDict(src);
+            }
+
+            std::string Decompress(const char *src, size_t srcSize, const bool useDict)
```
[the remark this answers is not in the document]
suiguoxin: Done, thanks

AnnService/inc/Core/SPANN/ExtraFullGraphSearcher.h
suiguoxin: Fixed by adding options instead of dumping configurations to head info. The head info format will only be changed when compression enabled.

AnnService/SSDServing.vcxproj
```diff
@@ -54,7 +54,7 @@
     <CharacterSet>MultiByte</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
-    <ConfigurationType>StaticLibrary</ConfigurationType>
+    <ConfigurationType>Application</ConfigurationType>
```
PhilipBAdams, line 57: Please revert, needs to be .lib for the tests to run

PhilipBAdams (approved):

PhilipBAdams:
Please also change SPTAG.nuspec to include your new files, it is what we use to generate the nuget package

AnnService/SSDServing.vcxproj
suiguoxin: removed

suiguoxin:
> Please also change SPTAG.nuspec to include your new files, it is what we use to generate the nuget package

Added. Is there something to do with submodule files?

PR #300, opened 2022-06-06
fix some bugs in BKTree and SPANNIndex, add new result format for indexsearcher

MaggieQi:


AnnService/src/IndexSearcher/main.cpp
```diff
@@ -69,6 +70,8 @@ class SearcherOptions : public Helper::ReaderOptions
     int m_debugQuery = -1;
 
     bool m_enableADC = false;
+
+    int m_outputformat = 0;
```
PhilipBAdams, line 74: Can we use SPTAG::VectorFileType instead

AnnService/src/IndexSearcher/main.cpp
```diff
@@ -237,37 +254,76 @@ int Process(std::shared_ptr<SearcherOptions> options, VectorIndex& index)
...
+                for (SizeType i = 0; i < numQuerys; i++)
                 {
-                    if (results[i].GetResult(j)->VID < 0) {
-                        fp << results[i].GetResult(j)->Dist << "@NULL" << std::endl;
-                        continue;
-                    }
-
-                    if (!options->m_withMeta) {
-                        fp << (results[i].GetResult(j)->Dist / baseSquare) << "@" << results[i].GetResult(j)->VID << std::endl;
+                    if (queryMetas != nullptr) {
```
PhilipBAdams, line 262: If we want to support multiple output formats, should we wrap it in a `VectorSetWriter` the way we do with multiple input formats? I think we have this kind of logic in SSDServing too

PhilipBAdams (approved):

AnnService/src/IndexSearcher/main.cpp
MaggieQi: It is the result output type not the input vector type

AnnService/src/IndexSearcher/main.cpp
MaggieQi: Yes, good suggestion! We can do it after this pull request.

PR #305, opened 2022-06-15
Support inplace sum operation with SIMD

suiguoxin:


suiguoxin:
> Looks good to me, but can we split it to another file? DistanceUtils is already large and vector sum is not a distance function.

@PhilipBAdams  Moved to SIMDUtils/SIMDTest

PhilipBAdams (approved):

Issue #327, opened 2022-07-31
print reason for file handle creation failure


pabs3:
In [AsyncFileReader.h](https://github.com/microsoft/SPTAG/blob/main/AnnService/inc/Helper/AsyncFileReader.h) class `AsyncFileIO` function `Initialize`, it handles failure to open the file by printing an error saying that the file handle couldn't be created, but it doesn't print the reason for the error. This can be printed by printing the string returned by a `strerror(errno)` function call.

pabs3:
This will be helpful to diagnose the [SPTAG build failures in Debian](https://buildd.debian.org/status/logs.php?pkg=sptag&arch=amd64).

PR #328, opened 2022-07-31
Print file open error information

pabs3:
Should help diagnose the build failures in Debian.

Fixes: https://github.com/microsoft/SPTAG/issues/327


pabs3:
I do not agree with the CLA.

I do not believe that these changes are copyrightable.

If they are copyrightable, I am the copyright holder and I release
these changes under the MIT license.

In both cases it should be possible to merge the changes.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #330, opened 2022-08-03
Separate head/ssd maxcheck, update nuspec

PhilipBAdams:


AnnService/src/Core/SPANN/SPANNIndex.cpp
```diff
@@ -113,7 +113,7 @@ namespace SPTAG
             if (m_index->LoadIndexData(p_indexStreams) != ErrorCode::Success) return ErrorCode::Fail;
 
             m_index->SetParameter("NumberOfThreads", std::to_string(m_options.m_iSSDNumberOfThreads));
-            m_index->SetParameter("MaxCheck", std::to_string(m_options.m_maxCheck));
+            //m_index->SetParameter("MaxCheck", std::to_string(m_options.m_maxCheck));
```
MaggieQi, line 116: why we need to separate these two maxcheck? If we would like to separate these two maxcheck, then we should also separate HashTableExponent since it is related to maxcheck.

MaggieQi (changes requested):

SPTAG.nuspec
```diff
@@ -78,6 +78,9 @@
     <file src="AnnService\inc\Helper\StringConvert.h" target="build\native\inc\Helper\StringConvert.h" />
     <file src="AnnService\inc\Helper\ThreadPool.h" target="build\native\inc\Helper\ThreadPool.h" />
     <file src="AnnService\inc\Helper\VectorSetReader.h" target="build\native\inc\Helper\VectorSetReader.h" />
+    <file src="ThirdParty\zstd\lib\zdict.h" target="build\native\zdict.h" />
+    <file src="ThirdParty\zstd\lib\zstd.h" target="build\native\zstd.h" />
+    <file src="ThirdParty\zstd\lib\zstd_errors.h" target="build\native\zstd_errors.h" />
     <file src="x64\Debug\CoreLibrary.lib" target="lib\native\debug\x64\CoreLibrary.lib" />
     <file src="x64\Debug\CoreLibrary.pdb" target="lib\native\debug\x64\CoreLibrary.pdb" />
     <file src="x64\Release\CoreLibrary.lib" target="lib\native\retail\x64\CoreLibrary.lib" />
```
MaggieQi, line 86: You will also need to pack the libzstd.lib and libzstd.dll into the nuget package.

AnnService/src/Core/SPANN/SPANNIndex.cpp
PhilipBAdams: Currently, we search up to `MaxCheck` in head index, then up to `MaxCheck` in SSD index, since they have different WorkSpace so the check count is not shared.

In my local test on small index, I can get better QPS by reduce the head maxcheck but keep SSD maxcheck the same. I'll add the `HashTableExponent` change.

SPTAG.nuspec
PhilipBAdams: We have internal zstd build, I will check if it is usable instead of packaging different version along w/ sptag

AnnService/src/Core/SPANN/SPANNIndex.cpp
```diff
@@ -658,7 +658,7 @@ namespace SPTAG
                 if (!CheckHeadIndexType()) return ErrorCode::Fail;
 
                 m_index->SetParameter("NumberOfThreads", std::to_string(m_options.m_iSSDNumberOfThreads));
-                m_index->SetParameter("MaxCheck", std::to_string(m_options.m_maxCheck));
+                //m_index->SetParameter("MaxCheck", std::to_string(m_options.m_maxCheck));
                 m_index->SetParameter("HashTableExponent", std::to_string(m_options.m_hashExp));
```
MaggieQi, line 662: similar separate the HashTableExponent.

PR #333, opened 2022-08-31
Fix python version

MaggieQi:


AnnService/inc/Core/Common.h
```diff
@@ -81,6 +81,15 @@ inline bool fileexists(const TCHAR* path) {
     return (dwAttr != INVALID_FILE_ATTRIBUTES) && (dwAttr & FILE_ATTRIBUTE_DIRECTORY) == 0;
 }
 #define mkdir(a) CreateDirectory(a, NULL)
+
+#ifndef max
+#define max(a,b) (((a) > (b)) ? (a) : (b))
```
PhilipBAdams, line 86: I think we should prefer to use `std::max`, define this macro can cause some conflicts with other code

AnnService/src/Helper/AsyncFileReader.cpp
```diff
@@ -93,33 +92,15 @@ namespace SPTAG {
...
-                
-                if (readRequest->m_success && readRequest->m_callback)
-                {
-                    readRequest->m_callback(true);
-                    readRequest->m_callback = nullptr;
+                    handlers[currFileId]->BatchReadFile(readRequests + currReqStart, num - currReqStart);
                 }
             }
-
-            CallOnAppropriateBatch(handlers, readRequests, num, &DiskIO::BatchCleanRequests);
```
PhilipBAdams, line 122: We still need to clean requests in windows build, to release the refcounted buffers in internal implementation

AnnService/inc/Core/Common.h
MaggieQi: this is required by dll compiling. It seems it cannot find the std::max.

AnnService/src/Helper/AsyncFileReader.cpp
MaggieQi: I have modified the DiskIO part so that let the diskio to clean the context instead of our library.

limenghao (approved):

PR #339, opened 2022-10-25
Fix low accuracy issue with GPU index build for int8/cosine configuration

bkarsin:
Fix for issue where GPU index build with int8 datatype and cosine distance metric resulting in low accuracy.  This is due to the base being incorrectly computed for int8.  Also enabled the hardware optimization for this case, which provides a ~60% perf gain over the standard cosine distance function.


MaggieQi (approved):
thanks Ben!

MaggieQi (approved):

PR #359, opened 2023-02-23
Thread local context

MaggieQi:


AnnService/inc/Core/BKT/Index.h
```diff
@@ -85,7 +84,8 @@ namespace SPTAG
             int m_iNumberOfInitialDynamicPivots;
             int m_iNumberOfOtherDynamicPivots;
             int m_iHashTableExp;
-
+        public:
+            static thread_local std::shared_ptr<COMMON::WorkSpace> m_workspace;
```
PhilipBAdams, line 88: I'm not sure it is safe to have workspace handled this way. If we have two objects index A and index B, both can run search on any thread. So if the config of the two index is different, it can cause problems or requires expensive reset each time workspace is used.

What is your opinion on alternative, have workspace passed in as argument and managed by the calling code? Or can we guarantee, even use each other's worspace, index A and B don't have any stability issue or other bug?

PhilipBAdams: I think the current incompatibilities are 

- Maxcheck (deduper config and queue sizes)
- data compression/pagebuffer size 
- p_InternalResultNum: Iocp size, posting sizes, etc.

AnnService/src/Helper/AsyncFileReader.cpp
```diff
@@ -6,6 +6,49 @@
...
 #ifndef _MSC_VER
+        void SetThreadAffinity(int threadID, std::thread& thread, char socketStrategy, char idStrategy)
+        {
+#ifdef NUMA
+            int numGroups = numa_num_task_nodes();
+            int numCpus = numa_num_task_cpus() / numGroups;
+
+            int group = threadID / numCpus;
+            int cpuid = threadID % numCpus;
+            if (socketStrategy == 1) {
```
PhilipBAdams, line 17: can we name these strategies by descriptive enum rather than int value?

AnnService/src/IndexSearcher/main.cpp
```diff
@@ -187,29 +188,31 @@ int Process(std::shared_ptr<SearcherOptions> options, VectorIndex& index)
...
-                    }
-                }
-            };
 
             auto batchstart = std::chrono::high_resolution_clock::now();
 
-            for (std::uint32_t i = 0; i < options->m_threadNum; i++) { threads.emplace_back(func); }
+            for (std::uint32_t i = 0; i < options->m_threadNum; i++) { 
+                threads.emplace_back([&, i] {
+                    Helper::SetThreadAffinity(i, threads[i], 0, 0);
```
PhilipBAdams, line 196: Should all calling code to SPTAG search, set it's affinity like this? Can we make some small doc or comment, explain the options and tradeoffs for SPTAG case?

CMakeLists.txt
```diff
@@ -26,6 +26,29 @@ if(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
...
+    if (NUMA_INCLUDE_DIR AND NUMA_LIBRARY AND NUMA_LIBRARY_STATIC)
+        set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -lnuma")
+        set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lnuma")
+
+        include_directories (${NUMA_INCLUDE_DIR})
+        message (STATUS "Found numa library: inc=${NUMA_INCLUDE_DIR}, lib=${NUMA_LIBRARY}, staticlib=${NUMA_LIBRARY_STATIC}")
+	set(Definition -DNUMA)
+    else ()
+        message (STATUS "WARNING: Numa library not found.")
+        message (STATUS "Try: 'sudo yum install numactl numactl-devel' (or sudo apt-get install libnuma libnuma-dev)")
```
PhilipBAdams, line 49: need to add these packages to Dockerfile, Dockerfile.cuda?

PhilipBAdams: Or, use cmake add_definitions to set NUMA macro rather than set as 1 in code

AnnService/inc/Core/BKT/Index.h
PhilipBAdams: If we are confident it's safe, let's prove it by unit test compare interleaved search on same thread of A and B to search on different threads. Then we can prevent regression if new feature adds more state to `workspace`.

MaggieQi: If different indices have different search thread pool, then we can ensure the stability.

AnnService/src/Helper/AsyncFileReader.cpp
MaggieQi: sure

AnnService/src/Core/BKT/BKTIndex.cpp
```diff
@@ -86,18 +87,16 @@ namespace SPTAG
...
             return ret;
         }
 
         template <typename T>
         ErrorCode Index<T>::SaveConfig(std::shared_ptr<Helper::DiskIO> p_configOut)
         {
-            auto workSpace = m_workSpacePool->Rent();
-            m_iHashTableExp = workSpace->HashTableExponent();
-            m_workSpacePool->Return(workSpace);
+            if (m_workspace.get() != nullptr) {
```
asklar, line 97: no need to call `.get()` here, you can just compare the smart pointer directly
(here and in other places)

CMakeLists.txt
```diff
@@ -26,6 +26,31 @@ if(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
...
+        set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lnuma")
+
+        include_directories (${NUMA_INCLUDE_DIR})
+        message (STATUS "Found numa library: inc=${NUMA_INCLUDE_DIR}, lib=${NUMA_LIBRARY}, staticlib=${NUMA_LIBRARY_STATIC}")
+        add_definitions(-DNUMA)
+    else ()
+        set (NUMA_LIBRARY "")
+        set (NUMA_LIBRARY_STATIC "")
+        message (STATUS "WARNING: Numa library not found.")
+        message (STATUS "Try: 'sudo yum install numactl numactl-devel' (or sudo apt-get install libnuma libnuma-dev)")
```
asklar, line 51: how would this work for NUMA on Windows? I don't see a reference to a numa nuget pkg or use of the Windows NUMA APIs?

PhilipBAdams: It's using the simpler processor-group apis, see AsyncFileReader.cpp:138

It might be worthwhile to use the numa apis explicitly if some cpu skus have >64 core in same node

PhilipBAdams (approved):

PR #360, opened 2023-02-24
Add WinRT wrapper

asklar:
Adds a WinRT wrapper and a test project

Still to come:
- nuspec
- pipeline update to publish the winrt package


Wrappers/WinRT/AnnIndex.idl
```diff
@@ -0,0 +1,37 @@
...
+  {
+    AnnIndex();
+    [default_overload]
+    void AddWithMetadata(Single[] data, Windows.Data.Json.JsonValue metadata);
+    void AddWithMetadata(Single[] data, Guid metadata);
+    void AddWithMetadata(Single[] data, String metadata);
+
+    void Save(Windows.Storage.StorageFile file);
+    void Load(Windows.Storage.StorageFile file);
+    Windows.Foundation.Collections.IVector<SearchResult> Search(Single[] vector, UInt8 neighborCount);
```
PhilipBAdams, line 35: This type should be wider?

Wrappers/WinRT/AnnIndex.idl
```diff
@@ -0,0 +1,37 @@
...
+    Error,
+    Assert,
+    Count,
+    Empty
+  };
+
+  [default_interface]
+  runtimeclass SearchResult
+  {
+    Windows.Data.Json.JsonValue Metadata {get; };
```
PhilipBAdams, line 19: Is it possible just have metadata as blob of bytes here? Algo does not consider the details, library consumer can pick whatever serialization they prefer

asklar: we could - though that means the onus is on the caller to serialize types, which as you can see isn't exactly easy :)
would it be ok to just add the overload for UInt8[] and let the user pick which data type they want to store?

PhilipBAdams (approved):

PR #385, opened 2023-05-22
Clr core

MaggieQi:


AnnService/src/Core/BKT/BKTIndex.cpp
```diff
@@ -328,7 +321,7 @@ namespace SPTAG
             bool NeverDup(COMMON::QueryResultSet<T>& query, SizeType node, float score)
             {
                 query.AddPoint(node, score);
-                return false;
+                return true;
```
PhilipBAdams, line 324: This will mean it will always break on the first iteration, I don't think it's correct

AnnService/inc/Core/BKT/Index.h
```diff
@@ -42,7 +42,26 @@ namespace SPTAG
                 RebuildJob(COMMON::Dataset<T>* p_data, COMMON::BKTree* p_tree, COMMON::RelativeNeighborhoodGraph* p_graph, 
                     DistCalcMethod p_distMethod) : m_data(p_data), m_tree(p_tree), m_graph(p_graph), m_distMethod(p_distMethod) {}
                 void exec(IAbortOperation* p_abort) {
-                    m_tree->Rebuild<T>(*m_data, m_distMethod, p_abort);
+                    COMMON::BKTree newTrees(*m_tree);
+                    newTrees.BuildTrees<T>(*m_data, m_distMethod, 1, nullptr, nullptr, false, p_abort);
+
+                    std::unique_lock<std::shared_timed_mutex> lock(*(m_tree->m_lock));
```
PhilipBAdams, line 48: Can build the new trees before acquire the lock cause data added after the start of building new trees but before the swap to be lost?

MaggieQi: We do not need to lock the data add since tree does not need to catch up with new data.

PhilipBAdams (approved):

PR #391, opened 2023-07-18
Handle checks for invalid deletedIDs

PhilipBAdams:
Avoid crashing during index merge when IDs from each index are duplicated to both indices, causing invalid IDs to be inserted

### DIFF
--- a/AnnService/inc/Core/Common/Labelset.h
+++ b/AnnService/inc/Core/Common/Labelset.h
@@ -13,19 +13,28 @@ namespace SPTAG
     {
         class Labelset
         {
+        public:
+            enum class InvalidIDBehavior
+            {
+                Passthrough,
+                AlwaysContains,
+                AlwaysNotContains
+            };
         private:
             std::atomic<SizeType> m_inserted;
             Dataset<std::int8_t> m_data;
+            InvalidIDBehavior m_invalidIDBehaviorSetting;
             
         public:
-            Labelset() 
+            Labelset()
             {
                 m_inserted = 0;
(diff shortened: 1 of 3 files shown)


AnnService/inc/Core/Common/Labelset.h
```diff
@@ -13,31 +13,62 @@ namespace SPTAG
...
 
             inline bool Contains(const SizeType& key) const
             {
+                if (key >= R())
+                {
+                    switch (m_invalidIDBehaviorSetting)
+                    {
+                        case InvalidIDBehavior::AlwaysContains:
+                            return true;
+                        case InvalidIDBehavior::AlwaysNotContains:
```
QingchaChen, line 51: need a default ?

AnnService/inc/Core/Common/Labelset.h
```diff
@@ -13,31 +13,62 @@ namespace SPTAG
...
             inline bool Insert(const SizeType& key)
             {
+                if (key >= R())
+                {
+                    switch (m_invalidIDBehaviorSetting)
+                    {
+                        case InvalidIDBehavior::AlwaysContains:
+                            return true;
+                        case InvalidIDBehavior::AlwaysNotContains:
+                            return false;
```
QingchaChen, line 68: need default

PhilipBAdams: the default case is Passthrough, it falls through to the old logic (dataset class throwing exception)

limenghao (approved):

PR #394, opened 2023-08-29
handle case where metadataset != vectorset size in index merge

PhilipBAdams:


AnnService/src/Core/BKT/BKTIndex.cpp
```diff
@@ -675,11 +675,15 @@ namespace SPTAG
                     if (p_metadataSet != nullptr) {
                         m_pMetadata->AddBatch(*p_metadataSet);
                         if (HasMetaMapping()) {
-                            for (SizeType i = begin; i < end; i++) {
+                            for (SizeType i = begin; i < end && i < begin + p_metadataSet->Count(); i++) {
```
MaggieQi, line 678: Hi Philip, why it will have the case where metadataset != vectorset size? Then how can you align vector and metadata?

MaggieQi (changes requested):

PR #407, opened 2023-12-20
Iterator interface

zqxjjj:
The iterator interface can get the result candidate one by one.
No strict monotonicity is in the results. 
When relaxedMono is set to true in BasicResult, it indicates that all data satisfying a certain accuracy level has been returned.


MaggieQi (dismissed):

MaggieQi (dismissed):
approved

MaggieQi (approved):

PR #423, opened 2025-06-04
Spfresh async io

MaggieQi:


PhilipBAdams (dismissed):

AnnService/src/Core/SPANN/SPANNIndex.cpp
```diff
@@ -416,15 +417,20 @@ ErrorCode Index<T>::SearchIndexIterative(QueryResult &p_headQuery, QueryResult &
...
         if (!found && ret != ErrorCode::VectorNotFound) return ret;
         p_extraWorkspace->m_loadPosting = false;
         if (!found)
         {
+            p_headQueryResults->SetResultNum(m_options.m_headBatch);
             p_headQueryResults->Reset();
             continueSearch = m_index->SearchIndexIterativeFromNeareast(*p_headQueryResults, p_indexWorkspace, false);
             p_extraWorkspace->m_loadPosting = true;
+
+            if (!oldRelaxedMono && p_extraWorkspace->m_relaxedMono)
```
PhilipBAdams, line 432: Note that this needs test change since we can return less than the batch argument, to give client a chance to decide whether to search further before doing more IO

PhilipBAdams (approved):

PR #448, opened 2026-05-07
[Feature] Add Distributed Posting Router for SPANN

TerrenceZhangX:
## Limitations
1. Using `buildOnly` flags to mitigate no ditributed build limitation. Current flow is using `buildOnly` to let single node to build and distribute to other nodes. Then using disable `buildOnly` to let multiple nodes load same head index and run benchmarks.
## Scale results

Dataset: SIFT1B `bigann_base.u8bin`, 128d UInt8, L2. SPANN 2-layer index,
4 search threads, 4 insert threads, top-K=5, 200 queries.

### 1M base + 1M insert

| Metric                       | 1node  | 2node  | Scale     |
| ---------------------------- | -----: | -----: | --------: |
| Build time (s)               |   74.2 |   91.3 | 0.81×     |
| Pre-insert QPS               |  429.3 |  696.3 | **1.62×** |
| Pre-insert mean latency (ms) |   9.26 |   8.63 |     —     |
| Pre-insert p99 (ms)          |  27.31 |  29.18 |     —     |
| Post-insert QPS              |  425.0 |  708.1 | **1.67×** |
| Insert throughput (vec/s)    |    ~900 |  1793 |     —     |
| Recall@5 (pre-insert)        |  0.984 |  0.978 |     —     |
| Recall@5 (post-insert)       |  0.983 |  0.984 |     —     |

Notes:
- Build is *slower* on 2node at 1M scale: the cross-node coordination
  overhead (head-sync RPCs, control plane) dominates when there are only
  ~40k head vectors to build.
- Search scales well already at 1M; recall is unchanged.

### 100M base + 1M insert

| Metric                       | 1node  | 2node   | Scale      |
| ---------------------------- | -----: | ------: | ---------: |
| Build time (s)               |  15292 |   16264 | 0.94×      |
| Pre-insert QPS               |  183.3 |   360.5 | **1.97×**  |
| Pre-insert mean latency (ms) |  21.56 |   21.92 |     —      |
| Pre-insert p99 (ms)          |  32.81 |   39.55 |     —      |
| Post-insert QPS (round 2)    |  183.2 |   337.2 | **1.84×**  |
| Insert throughput (vec/s)    |    738 |    1285 | **1.74×**  |
| Recall@5 (pre-insert)        |  0.912 |   0.904 |     —      |
| Recall@5 (post-insert)       |  0.912 |   0.904 |     —      |

Notes:
- **Search scales near-linearly (1.97×)** at the target scale — the per-node
  query partition + remote KV reads are well-balanced.
- **Insert scales sublinearly (1.74×)**, expected: every insert that
  promotes/splits a head triggers head-index sync across nodes, which is the
  current bottleneck.
- Build is essentially flat (0.94×): build is dominated by per-node local
  graph construction; 2node has additional head-sync but no work split, so
  near-no scaling here is expected for the current single-builder design.
- Recall is stable across configurations (within 0.01).


AnnService/inc/Core/Common/FineGrainedLock.h
```diff
@@ -56,9 +56,13 @@ namespace SPTAG
...
             }
 
+            // Public hash used by external callers (e.g. ExtraDynamicSearcher's
+            // m_remoteBucketLocked array, sized kRemoteLockPoolSize+1 == BucketCount)
+            // to derive a bucket index that matches the granularity of this lock pool.
+            // MUST stay in sync with BucketIndex below.
             static inline unsigned hash_func(unsigned idx)
             {
-                return idx;
+                return ((unsigned)(idx * 99991) + _rotl(idx, 2) + 101) & BucketMask;
```
MaggieQi: Here we switch to per posting lock, therefore, hash function directly returns idx

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -2048,33 +3158,122 @@ namespace SPTAG::SPANN {
...
+                    SizeType deficit = maxVid - m_versionMap->GetVectorNum() + 1;
+                    int numWorkers = std::max(1, GetNumWorkerNodes());
+                    int64_t safeBound = std::max<int64_t>((int64_t)1 << 20,
+                        (int64_t)currentCount * std::max(numWorkers, 2) + ((int64_t)1 << 24));
+                    SizeType maxAllowed = (SizeType)std::min<int64_t>(safeBound, (int64_t)INT_MAX >> 1);
+                    if (deficit > maxAllowed) {
+                        SPTAGLIB_LOG(Helper::LogLevel::LL_Error,
+                            "[Bug 25] Search post-fetch: out-of-range VID:%d (count:%d, deficit:%d > %d, numWorkers:%d); records will be skipped\n",
+                            maxVid, currentCount, deficit, maxAllowed, numWorkers);
+                    } else if (m_versionMap->AddBatch(deficit) != ErrorCode::Success) {
```
MaggieQi: why search need to be changed?

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -2108,7 +3307,32 @@ namespace SPTAG::SPANN {
...
+                //     saturating m_splitThreadPool with noop work and
+                //     starving real layer-0 split/append work.
+                // (b) Only trigger when this head is locally owned. The
+                //     Bug-11 fast path in MergePostings already drops
+                //     remote-owned merges, but the job submission/scheduling
+                //     overhead is still ~50% of the total cost. Filtering
+                //     at the trigger site avoids the round trip entirely.
+                // Single-node behavior (m_worker disabled) is preserved.
+                if (m_opt->m_asyncMergeInSearch && realNum <= m_mergeThreshold) {
+                    if (m_worker && m_worker->IsEnabled()) {
```
MaggieQi: Even though it is not local and not layer 0, it still needs to do mergeasync.

MaggieQi (changes requested):

AnnService/inc/Core/Common/FineGrainedLock.h
TerrenceZhangX: You're right. The intent was a per-bucket lock pool for the remote RPC path, but BucketIndex was private, so Copilot misused the public hash_func (which is per-posting identity) as a bucket hash. Now BucketIndex is exposed and the remote lock pool uses it directly; hash_func is back to identity for the self-deadlock guards.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: Each node's local m_count tracks only its own AddIndex progress, not peers'. When search reads a TiKV posting whose bytes were written by a peer running ahead of us, those record VIDs can fall above our local m_count. GetVersion short-circuits any VID >= m_count to 0xfe, so without growing m_count first the post-heap filter would drop those records as deleted and recall would silently regress. This loop grows m_count just enough to cover the largest VID we actually fetched

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: Agree. This is over-simplification. Fixed to mergeAsync trigger based on headID owner.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -339,9 +354,244 @@ namespace SPTAG::SPANN {
...
+            }
+            req.m_appendNum = appendNum;
+            req.m_appendPosting = std::move(posting);
+            m_worker->QueueRemoteAppend(nodeIndex, std::move(req));
+        }
+
+        // If headID is owned by a remote node, queue the append for that
+        // node and return true; otherwise return false (caller continues
+        // with local write logic).
+        bool TryRouteRemoteAppend(SizeType headID,
```
TerrenceZhangX: Can we find somewhere to just identify once?

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -925,20 +1208,36 @@ namespace SPTAG::SPANN {
...
-                                return ret;
+                            // If newHeadVID's owner is a remote node, route
+                            // the initial posting via RemoteAppend so it
+                            // ends up in the owner's TiKV. We still add the
+                            // head locally and rely on BroadcastHeadSync
+                            // (after this loop) to spread the head index
+                            // update to all nodes. The receiver's
+                            // AppendCallback materializes the head if its
+                            // HeadSync hasn't arrived yet.
+                            bool remoteCreated = TryRouteRemoteAppend(
```
TerrenceZhangX: Should be two locks get together then do the job

TerrenceZhangX: Certain fault tolerance is also needed

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -962,6 +1261,35 @@ namespace SPTAG::SPANN {
                     }
                 }
```
TerrenceZhangX, line 2055: On machine A: S split to S1 and S2

On machine B: S1 splits into S2 and S3. Will there be issue?

1. S-> S1/S2
2. S1 -> S2/S3
3. BC S1-> S2/S3
4. BC S-> S1/S2. Issue: S1 is back

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -1003,6 +1331,18 @@ namespace SPTAG::SPANN {
 
         ErrorCode MergePostings(ExtraWorkSpace *p_exWorkSpace, SizeType headID)
         {
+            // The owner runs its own merge passes. Skip when this head is
+            // owned by another node — we'd just be racing the owner.
+            if (m_worker && m_worker->IsEnabled()) {
```
TerrenceZhangX: If refine is not there, do we still need the filter

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -1102,23 +1442,61 @@ namespace SPTAG::SPANN {
...
+                                m_splitThreadPool->add(curJob);
+                                return ErrorCode::Success;
+                            }
                         }
+                        if (!m_headIndex->ContainSample(queryResult->VID, m_layer + 1)) continue;
                     }
-                    if (!m_headIndex->ContainSample(queryResult->VID, m_layer + 1)) continue;
+
                     if ((ret=db->Get(DBKey(queryResult->VID), &nextPostingList, MaxTimeout, &(p_exWorkSpace->m_diskRequests))) != ErrorCode::Success) {
+                        if (isRemoteCandidate) {
```
TerrenceZhangX: is this correct?

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -1158,11 +1544,25 @@ namespace SPTAG::SPANN {
                             return ret;
                         }
                         CheckCentroid(headID, mergedPostingList, "MergePostings-currentLength >= nextLength");
-                        m_headIndex->DeleteIndex(queryResult->VID, m_layer + 1);
-                        if ((ret=db->Delete(DBKey(queryResult->VID))) != ErrorCode::Success)
-                        {
-                            SPTAGLIB_LOG(Helper::LogLevel::LL_Error, "Fail to delete old posting %lld in Merge\n", (std::int64_t)(queryResult->VID));
-                            return ret;
+                        if (isRemoteCandidate) {
```
TerrenceZhangX: no lock and delete?

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -2556,10 +3061,20 @@ namespace SPTAG::SPANN {
...
+                    if (shared) {
+                        m_splitThreadPool = std::static_pointer_cast<SPDKThreadPool>(shared);
+                    }
+                }
+                if (m_splitThreadPool == nullptr) {
+                    SPTAGLIB_LOG(Helper::LogLevel::LL_Info, "SPFresh: initialize thread pools, append: %d, reassign %d\n", m_opt->m_appendThreadNum, m_opt->m_reassignThreadNum);
+                    m_splitThreadPool = std::make_shared<SPDKThreadPool>();
+                    m_splitThreadPool->initSPDK(m_opt->m_appendThreadNum, this);
+                    SPTAGLIB_LOG(Helper::LogLevel::LL_Info, "SPFresh: finish initialization, zeroReplicaCount:%zu\n", zeroReplicaCount);
+                    if (m_layer == 0 && m_headIndex) m_headIndex->SetSharedSplitPool(m_splitThreadPool);
```
TerrenceZhangX: every layer has its split pool.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: Known issue. Cannot solve.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: fixed

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -206,16 +259,39 @@ namespace SPTAG::SPANN {
             }
         };
 
+    public:
+        // Expose the underlying KV handle so a standalone WorkerNode can be wired to the
+        // same DB this searcher already opened, instead of opening a second one.
+        std::shared_ptr<Helper::KeyValueIO> GetDB() const { return db; }
+
     private:
+        std::atomic<int> m_workspaceCount = 0;
```
MaggieQi: m_workspacecount is not used in this file.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -206,16 +259,39 @@ namespace SPTAG::SPANN {
...
         std::shared_ptr<Helper::KeyValueIO> db;
+        WorkerNode* m_worker = nullptr;  // externally owned, set via SetWorker()
 
         SPANN::Index<ValueType>* m_headIndex;
         std::unique_ptr<COMMON::IVersionMap> m_versionMap;
         Options* m_opt;
         int m_layer;
+        SizeType m_initialVectorSize = 0;  // vector count at build time (before inserts)
```
MaggieQi: total vector count? Why not write to/read from db?

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: deleted

AnnService/src/Socket/Connection.cpp
```diff
@@ -26,10 +26,19 @@ Connection::Connection(ConnectionID p_connectionID, boost::asio::ip::tcp::socket
...
-                 m_socket.remote_endpoint().address().to_string().c_str(),
-                 static_cast<uint32_t>(m_socket.remote_endpoint().port()));
+    boost::system::error_code epEc;
+    auto localEp = m_socket.local_endpoint(epEc);
+    auto remoteEp = m_socket.remote_endpoint(epEc);
+    if (!epEc) {
+        SPTAGLIB_LOG(Helper::LogLevel::LL_Debug, "Connection Start, local: %u, remote: %s:%u\n",
+                     static_cast<uint32_t>(localEp.port()),
+                     remoteEp.address().to_string().c_str(),
+                     static_cast<uint32_t>(remoteEp.port()));
```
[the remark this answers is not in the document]
TerrenceZhangX: Fixed in 047ed5b5. `Start()` now uses two separate `boost::system::error_code` instances (`localEc`, `remoteEc`), only logs the success branch when both calls succeeded, and the failure branch now reports which side errored. Same fix applied to `Stop()` in the same commit.

AnnService/src/Socket/Connection.cpp
```diff
@@ -42,10 +51,15 @@ void Connection::Start()
...
-                 static_cast<uint32_t>(m_socket.remote_endpoint().port()));
+    boost::system::error_code epEc;
+    auto localEp = m_socket.local_endpoint(epEc);
+    auto remoteEp = m_socket.remote_endpoint(epEc);
+    if (!epEc) {
+        SPTAGLIB_LOG(Helper::LogLevel::LL_Debug, "Connection Stop, local: %u, remote: %s:%u\n",
+                     static_cast<uint32_t>(localEp.port()),
+                     remoteEp.address().to_string().c_str(),
+                     static_cast<uint32_t>(remoteEp.port()));
+    }
```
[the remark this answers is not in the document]
TerrenceZhangX: Fixed in 047ed5b5 together with the matching Start() fix. `Stop()` now uses two separate `error_code` instances so the diagnostic log accurately reflects whichever endpoint actually errored.

AnnService/inc/Core/SPANN/Distributed/WorkerNode.h
```diff
@@ -0,0 +1,707 @@
...
+            // receiver runs 8 worker threads on each chunk that already
+            // interleave items within a chunk, so cross-chunk ordering adds
+            // no extra correctness risk for the per-posting RMW path.
+            auto items = std::make_shared<std::vector<RemoteAppendRequest>>(std::move(toFlush));
+            m_inflightAppendFlushes.fetch_add(1, std::memory_order_relaxed);
+            std::thread([this, nodeIndex, items]() {
+                while (true) {
+                    ErrorCode ret = SendBatchRemoteAppend(nodeIndex, *items);
+                    if (ret != ErrorCode::Success) {
+                        // Hand the failed batch to the watchdog. It owns
```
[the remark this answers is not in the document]
TerrenceZhangX: Fixed in 149bdd4d. Added `~WorkerNode()` with a gate-then-drain shutdown:

1. `m_acceptingNewRequests` is set to false. `QueueRemoteAppend` checks this at the top and returns early with a warning so no new auto-flush thread can spawn.
2. Destructor waits unbounded on `m_inflightAppendFlushes == 0`. With the gate set, each in-flight detached thread is bounded by one `SendBatchRemoteAppend` cycle (~one RPC timeout) plus one final loop iteration that sees an empty queue and breaks — so worst-case wall time is one RPC timeout regardless of concurrency. A periodic `LL_Warning` at 2× RPC timeout flags wedged-gRPC cases without false-alarming healthy drains.
3. Members tear down after the wait — `m_asyncWatchdog` then joins its own loop thread, then the mutex / queue / etc. destruct with no live consumers.

A hard timeout was considered and rejected: breaking out early would let detached threads outlive the captured members and immediately UAF — strictly worse than a slow shutdown.

Note: remote-node-failure was never a UAF for this path — `SendBatchRemoteAppend` to a dead peer goes through the local `m_asyncWatchdog` retry loop and gives up after `MaxAttempts`; the hazard was purely tied to local destruction.

AnnService/inc/Core/SPANN/Distributed/SplitWAL.h
```diff
@@ -0,0 +1,108 @@
...
+        Stage         stage;
+
+        std::string Encode() const {
+            std::string s(sizeof(Record), '\0');
+            memcpy(&s[0], this, sizeof(Record));
+            return s;
+        }
+        bool Decode(const std::string& s) {
+            if (s.size() < sizeof(Record)) return false;
+            memcpy(this, s.data(), sizeof(Record));
```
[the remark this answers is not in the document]
TerrenceZhangX: Fixed in 1786092d. Replaced `memcpy(this, ..., sizeof(Record))` with explicit field-wise serialization at known offsets and exposed `kEncodedSize` as a constexpr.

Wire layout (in order, no padding):
- uint64 jobID
- SizeType srcHeadID / localChildHeadID / remoteChildHeadID
- int    remoteOwnerNodeIndex
- int64  startTimestampSec
- uint8  stage

This eliminates padding leak, makes the encoding stable against source-order changes, and pins `stage` to a fixed `std::uint8_t` regardless of the enum's underlying-type declaration. SizeType width still tracks the build-config `LARGEVID` flag by design — the WAL contract is that writer and reader must be built with the same setting, matching the rest of the on-disk format.

AnnService/inc/Socket/SimpleSerialization.h
```diff
@@ -82,6 +82,58 @@ namespace SimpleSerialization
...
+    /// Bounds-checked variants of SimpleReadBuffer.
+    /// All return nullptr if a read would overrun [p_buffer, p_bufEnd).
+    /// p_buffer is also returned as nullptr (and p_val left unchanged) if it is already nullptr.
+    template<typename T>
+    inline const std::uint8_t*
+    SafeSimpleReadBuffer(const std::uint8_t* p_buffer, const std::uint8_t* p_bufEnd, T& p_val)
+    {
+        static_assert(std::is_fundamental<T>::value || std::is_enum<T>::value,
+                      "Only applied for fundanmental type.");
+
```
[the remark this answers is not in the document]
TerrenceZhangX: Fixed in 82dc35a8. Replaced all four `fundanmental` occurrences with `fundamental` in `SimpleWriteBuffer` / `SimpleReadBuffer` / `EstimateBufferSize` / `SafeSimpleReadBuffer`.

Test/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ if (NOT LIBRARYONLY)
     file(GLOB TEST_HDR_FILES ${PROJECT_SOURCE_DIR}/Test/inc/Test.h)
     file(GLOB TEST_SRC_FILES ${PROJECT_SOURCE_DIR}/Test/src/*.cpp)
     add_executable(SPTAGTest ${TEST_SRC_FILES} ${TEST_HDR_FILES})
-    target_link_libraries(SPTAGTest SPTAGLibStatic ssdservingLib ${Boost_LIBRARIES})
+    target_link_libraries(SPTAGTest SPTAGLibStatic ssdservingLib ${Boost_LIBRARIES} ${TiKV_LIBRARIES} absl_synchronization absl_cord absl_cordz_info absl_cord_internal absl_cordz_functions absl_cordz_handle)
```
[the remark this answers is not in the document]
TerrenceZhangX: Fixed in 8fd4c308. The `absl_*` link deps are now nested inside an `if (TIKV)` block, matching the convention used by `TiKV_LIBRARIES` in the top-level `CMakeLists.txt` (option declared at L131, libs gated at L172-201). When `TIKV=OFF` (the default) the absl libs are no longer required, restoring upstream's link line for non-TiKV builds.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -665,29 +1263,59 @@ namespace SPTAG::SPANN {
...
+                    // In distributed mode the version map's Count() is a per-node
+                    // local atomic. Global VIDs are striped across nodes, so a
+                    // vector owned/inserted by another node (global VID >= local
+                    // Count()) can be legitimately remote-appended into this
+                    // node's posting without growing the local count. Only treat
+                    // VID < 0 (a torn/garbage read) as corruption there; the
+                    // upper-bound check is single-node only. Downstream
+                    // BatchGetVersions handles out-of-range VIDs safely.
+                    bool distributed = (m_worker && m_worker->IsEnabled());
+                    SizeType maxVid = m_versionMap->Count();
```
MaggieQi: now versionMap is a real map, the count cannot represent the maxVID. It is better to have a maxVID stored in DB and get each time needed.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -766,29 +1394,244 @@ namespace SPTAG::SPANN {
...
+                            int owner = -1;
+                            if (IsRemoteOwnedHead(plans[k].newHeadVID, &owner)) {
+                                plans[k].isRemote = true;
+                                plans[k].ownerNode = owner;
+                            }
+                        }
+                    }
+                }
+
+                // === Phase B: build per-child posting payloads (memory only) ===
```
MaggieQi: why not combine phase A & B?

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -956,6 +1912,43 @@ namespace SPTAG::SPANN {
                     }
                 }
 
+                // Broadcast HeadSync to peer nodes when the head update lands
+                // in our local BKT (in-memory, per-compute). Lower-layer head
+                // adds that resolve to m_extraSearchers[m_layer+1]->AddIndex
+                // already write to shared TiKV so re-broadcasting them would
+                // only duplicate.
+                if (m_worker && m_worker->IsEnabled()
+                    && m_headIndex->GetDiskIndex(m_layer + 1) == nullptr) {
```
MaggieQi, line 2062: head in memory index should routing to one leader node for modification. Otherwise there are many nodes will do head modifications and overwrite each other.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -1090,40 +2111,68 @@ namespace SPTAG::SPANN {
...
+                        if (!remoteLease.acquire(m_worker, remoteNodeIndex, m_layer, queryResult->VID)) {
+                            continue;
+                        }
+                    } else {
+                        if (m_rwLocks.hash_func(queryResult->VID) != m_rwLocks.hash_func(headID)) {
+                            if (!anotherLock.try_lock()) {
+                                continue;
+                            }
                         }
+                        if (!m_headIndex->ContainSample(queryResult->VID, m_layer + 1)) continue;
```
MaggieQi: this line should be put in 2138?

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -1358,18 +2456,29 @@ namespace SPTAG::SPANN {
...
-                for (int j = 0; j < postVectorNum; j++) {
+                // Batched version-byte read for the entire posting.
+                std::vector<SizeType> cr_vids;
+                cr_vids.reserve(postVectorNum);
+                for (size_t j = 0; j < postVectorNum; j++) {
+                    cr_vids.push_back(*((SizeType*)(postingP + j * m_vectorInfoSize)));
+                }
+                std::vector<uint8_t> cr_mapVers;
+                m_versionMap->BatchGetVersions(cr_vids, cr_mapVers);
+                const SizeType maxVid = m_versionMap->Count();
```
MaggieQi: maxId is not correct now.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -1434,18 +2543,29 @@ namespace SPTAG::SPANN {
...
-                    for (int j = 0; j < postVectorNum; j++) {
+                    // Batched version-byte read for the nearby posting.
+                    std::vector<SizeType> nb_vids;
+                    nb_vids.reserve(postVectorNum);
+                    for (size_t j = 0; j < postVectorNum; j++) {
+                        nb_vids.push_back(*((SizeType*)(postingP + j * m_vectorInfoSize)));
+                    }
+                    std::vector<uint8_t> nb_mapVers;
+                    m_versionMap->BatchGetVersions(nb_vids, nb_mapVers);
+                    const SizeType maxVid = m_versionMap->Count();
```
MaggieQi: maxVId is not correct now

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -1795,7 +2987,9 @@ namespace SPTAG::SPANN {
 
         bool LoadIndex(Options& p_opt) override {
             m_opt = &p_opt;
-            SPTAGLIB_LOG(Helper::LogLevel::LL_Info, "DataBlockSize: %lld, Capacity: %lld\n", (std::int64_t)(m_opt->m_datasetRowsInBlock), (std::int64_t)(m_opt->m_datasetCapacity));
+            m_initialVectorSize = p_opt.m_vectorSize;  // initial count for VID stripe
```
MaggieQi: this is better to sync from db.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -2262,6 +3459,7 @@ namespace SPTAG::SPANN {
             {
                 auto fullVectors = p_reader->GetVectorSet();
                 fullCount = fullVectors->Count();
+                m_initialVectorSize = fullCount;  // remember bulk-build count for stripe formula
```
MaggieQi: load from db is better

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
```diff
@@ -2457,19 +3655,26 @@ namespace SPTAG::SPANN {
...
+            //     MergePostings reads stored version=0xfe, sees Deleted()=true
+            //     (because per-VID byte is missing → reads default 0xfe),
+            //     filters every entry, and writes back a corrupted near-empty
+            //     posting -- destroying recall after even a single async merge.
+            // The alive marker is 0xff to match the unified version convention
+            // (0xff = alive/default/reset, 0xfe = deleted) introduced on the
+            // pre-merge-tikv-bugfix branch; this is the batched-MultiPut
+            // equivalent of that branch's per-head
+            // "if (Deleted(globalID)) SetVersion(globalID, 0xff)" loop.
+            m_versionMap->Initialize(m_opt->m_vectorSize,
```
MaggieQi: Don't need to Init now.

MaggieQi: Using below code to init:
            if (p_localToGlobal.R() > 0) {
                for (SizeType i = 0; i < p_localToGlobal.R(); i++) {
                    SizeType globalID = *(p_localToGlobal[i]);
                    if (m_versionMap->Deleted(globalID)) m_versionMap->SetVersion(globalID, 0xff);
                }
            } else {
                for (SizeType i = 0; i < m_opt->m_vectorSize; i++) {
                    if (m_versionMap->Deleted(i)) m_versionMap->SetVersion(i, 0xff);
                }
            }

MaggieQi (changes requested):

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: Yes, this is the global total vector count after build. Renamed to m_initialGlobalTotalVectorCount to avoid misleading and change to write to/read from db.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: Agree. Deleted

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: combined.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: yes

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: fixed

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: fixed

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: fixed

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: Fixed.  head modifications are now routed to the head's owner node.
The fix, all in  ExtraDynamicSearcher.h :
•  PublishHeadSync()  (:882) partitions entries by  IsRemoteOwnedHead() . Locally-owned entries take the original path (append to our own shard log + broadcast). Peer-owned entries are stamped  m_shard = -1  and shipped to the owner via  WorkerNode::SendHeadSyncTo()  ( WorkerNode.h:290 ).
•  m_shard < 0  on the wire means "you are the owner, please publish this". The HeadSync callback (:713) branches on it into  HandleHeadSyncPublishRequest()  (:949), which re-checks ownership, applies locally, then publishes from its own shard log.
• Both publish sites — Split (:2085) and Merge (:2448) — now go through  PublishHeadSync()  instead of appending to the log directly.

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: fixed. All such sites now use  m_versionMap->MaxVID() , which for the TiKV-backed map reads the shared  vm:{layer}  watermark from the DB ( TiKVVersionMap.h:371 )

AnnService/inc/Core/SPANN/ExtraDynamicSearcher.h
TerrenceZhangX: fixed. All such sites now use  m_versionMap->MaxVID() , which for the TiKV-backed map reads the shared  vm:{layer}  watermark from the DB ( TiKVVersionMap.h:371 )

PR #451, opened 2026-06-10
SPANN: opt-in mmap of the input vector file

fastio:
## Motivation

`SPANN BuildIndex` currently reads the whole input vector file into heap before
SelectHead. For billion-scale inputs (e.g. SPACEV1B) that means hundreds of GB
of resident memory just to hold the raw vectors, on top of the build's own
working set. BKT clustering touches vectors randomly by id and rarely needs
them all hot at once.

## Design

A new `[Base] MmapVectors` option (default `false`). When enabled and the input
is the DEFAULT binary format, `DefaultVectorReader::GetVectorSet` memory-maps
the file read-only instead of reading it:

- The OS demand-pages vectors on access and reclaims clean file-backed pages
  under memory pressure, so peak RSS stays bounded.
- The mapping is wrapped in a small header-only RAII class
  (`Helper::MemoryMappedFile`, Windows + POSIX). Its lifetime follows the
  returned `VectorSet` via an aliasing `shared_ptr`, so the mapping is unmapped
  exactly once, after the last `ByteArray` copy is gone.
- `MADV_RANDOM` is applied on POSIX to suppress readahead, matching BKT's
  random access pattern.

### Fallback and safety

- If mapping fails for any reason, the reader falls back to
  in-memory read with a warning.
- The mapping is read-only, but SelectHead normalizes vecto
  the metric is Cosine and the input is not pre-normalized. `BuildIndex`
  detects that combination, disables mmap, and logs a warni
  to pre-normalize or use L2. L2 and pre-normalized Cosine inputs use mmap
  unmodified.


MaggieQi (approved):

fastio:
@microsoft-github-policy-service agree

MaggieQi (approved):
trigger ci build

